PR #1, opened 2024-07-10
FS-4529 initial db

srh-sloan:
https://dluhcdigital.atlassian.net/browse/FS-4529
https://dluhcdigital.atlassian.net/browse/FS-4530

Refactoring of the FAB prototype to include DB entities. Model classes, migrations, and a usable (not designed!) UI to view this data. 

To run locally, follow the instructions in the README. Recommend using the devcontainer setup, and then creating a docker compose file to start up this and the form runner.

Have added notes to the readme, but the stuff circled in red on the below screenshot is what's new in this PR and uses the database. Everything above that is the old drag and drop prototype, which has not been fully updated to use the DB so is not completely working at present. This will be fixed in later PRs as more functionality is updated to use the DB.

Screenshot 2024-07-10 at 11 56 20


app/blueprints/fund_builder/templates/round.html
```diff
@@ -0,0 +1,39 @@
...
+                          "text": form.fund_id.label
+                        },
+                        "items": all_funds})}}
+                    {{input(form.title_en)}}
+                    {{input(form.short_name)}}
+                    {{input(form.opens)}}
+                    {{input(form.deadline)}}
+                    {{input(form.assessment_start)}}
+                    {{input(form.reminder_date)}}
+                    {{input(form.assessment_deadline)}}
```
Adam-W1, line 27: not sure how finalised this one is to be but hints here for date format would be useful (or even a datetime field input:https://design-system.service.gov.uk/components/date-input/ )

app/db/models/application_config.py
```diff
@@ -0,0 +1,195 @@
...
+        UUID(as_uuid=True),
+        ForeignKey("theme.theme_id"),
+        nullable=True,  # will be null where this is a template and not linked to a theme
+    )
+    # TODO make these 2 json so we can do welsh?
+    title = Column(String())
+    hint_text = Column(String(), nullable=True)
+    options = Column(JSON(none_as_null=False))
+    type = Column(ENUM(ComponentType))
+    template_name = Column("Template Name", String(), nullable=True)
```
Adam-W1, line 159: I thought we might of had a conflict with the Third Normal form here
**
Transitive Dependency: This occurs when one non-key column is functionally dependent on another non-key column, which itself is dependent on the primary key.**

I dont think this is the case though, as everything appears to point back to the PK (unless I have misunderstood).

app/db/models/application_config.py
```diff
@@ -0,0 +1,195 @@
...
+    is_template = Column("is_template", Boolean, default=False, nullable=False)
+    audit_info = Column("audit_info", JSON(none_as_null=True))
+    page_index = Column(Integer())
+    theme_index = Column(Integer())
+    conditions = Column(JSON(none_as_null=True))
+    source_template_id = Column(UUID(as_uuid=True), nullable=True)
```
Adam-W1, line 165: is it worth having some sort of integrity so we cant inadvertently delete a source template (assuming this link is needed for some reason), for example

```suggestion
CREATE TABLE TemplatesTable (
    template_id INT PRIMARY KEY,
    source_template_id INT,
    -- other columns
    FOREIGN KEY (source_template_id) REFERENCES TemplatesTable(template_id) ON DELETE RESTRICT
);
```

app/db/models/application_config.py
```diff
@@ -0,0 +1,195 @@
...
+    )
+    name_in_apply = Column("name_in_apply_json", JSON(none_as_null=True), nullable=False, unique=False)
+    template_name = Column("Template Name", String(), nullable=True)
+    is_template = Column("is_template", Boolean, default=False, nullable=False)
+    audit_info = Column("audit_info", JSON(none_as_null=True))
+    forms: Mapped[List["Form"]] = relationship(
+        "Form", order_by="Form.section_index", collection_class=ordering_list("section_index")
+    )
+    index = Column(Integer())
+    source_template_id = Column(UUID(as_uuid=True), nullable=True)
```
Adam-W1, line 60: Just thinking, anything below this (form/component) must also be a template, as 'copying' a template will require a copy of all the tree below. Do we need to ensure that if a form is a template and it is FK to a section that both are template? Happy to discuss if that isnt clear

Adam-W1:
lovely work @srh-sloan as always.

Just left a few comments and possibly talking point. 

Rather that comment on page there are just a few other UI things that would be nice:
- Hint on data inputs
- Specific field inputs (datetime)
-Implement back button on journey
- CSS issue on button placement
- Static file loading (as discussed this morning on slack)
- A way of deleting sections/forms etc
- 
I realise these may be covered in other tickets that i havnt seen

app/blueprints/fund_builder/templates/round.html
srh-sloan: Yes totally. Very much just a dev-usage page atm. Didn't spend too much time on UI until mika has designed the proper journey

app/db/models/application_config.py
srh-sloan: Possibly. I think we were keeping it so that if a template changed, we could potentially in the future find out what was created from it and update those if needed (eg. a spelling mistake). It was more kind of crystal ball gazing than a concrete need. Don't think it matters though if the source template is deleted.

app/db/models/application_config.py
srh-sloan: Good question. I think that will come up in the next ticket around cloning from a template so will work out any specifics like this then if that's ok?

app/db/models/application_config.py
srh-sloan: Yeah this is just a 'display name' for when it's a template

PR #2, opened 2024-07-15
BAU renaming to FAB

srh-sloan:
Renaming from self-serve to FAB


Adam-W1:
Would you want to change all references to self-serve like those in app.py?

also i think this will need changing: https://github.com/communitiesuk/funding-service-design-fund-application-builder/blob/1c1c8ae6c53ab20a50b3d63fc8ea47a2bccbb6d0/app/app.py#L30

Adam-W1 (approved):
LGTM

PR #32, opened 2024-09-13
Fs 4606 use template

srh-sloan:
### Change description
Updated the import script to just import a template, not associate it with a round.
Enabled cloning a form template into a sepcific section
Updated clone functions to work with `default_next_page_id`

### How to test
Load up fab with a DB and form runner
As per the import README, import at least one form as a template
Navigate to the home screen, there's a new option 'Configure a round application'
Navigate through that option to where you can add a form to a section and do that
Then go to 'view fund config' and generate the export json for that section you've created (or preview the form)

### Scr
Screenshot 2024-09-13 at 12 56 20
eenshots of UI changes (if applicable)
Screenshot 2024-09-13 at 12 56 32


app/import_config/load_form_json.py
```diff
@@ -174,15 +173,16 @@ def insert_form_config(form_config, form_id):
...
 
 def insert_form_as_template(form):
-    section = db.session.query(Section.section_id).first()
+    #  section = db.session.query(Section.section_id).first()
+    form_name = next(p for p in form["pages"] if p["controller"] and p["controller"].endswith("start.js"))["title"]
     new_form = Form(
-        section_id=section.section_id,
-        name_in_apply_json={"en": form.get("name")},
-        template_name=form.get("name"),
+        section_id=None,  # ection.section_id,
```
Adam-W1, line 179: small typo

tests/test_clone.py
```diff
@@ -303,6 +304,32 @@ def test_clone_page_no_components(seed_dynamic_data, _db):
 page_id = uuid4()
 
 
+def test_fix_clone_default_pages():
+    # DISCLAIMER: Written on the train when I'd been up since 5am so may not be the most efficient/best approach!
```
Adam-W1, line 308: I guess you could clone a form here instead and then make sure that each page in the cloned form points to a cloned page id rather than the original?
If you just want to test the actual function here (probs what your going for tbf), I think it might be improved by specifying the original template pages and then the fixed cloned pages like so:

`
    original_pages = ..
`
`
    cloned_pages = .. (shows the error in default_page_id)
`
`
    fixed_pages = ..
`

app/db/queries/application.py
```diff
@@ -113,9 +127,25 @@ def clone_single_section(section_id: str, new_round_id=None) -> Section:
     return clone
 
 
-def clone_single_form(form_id: str, new_section_id=None) -> Form:
+def _fix_cloned_default_pages(cloned_pages: list[Page]):
```
Adam-W1, line 130: would be nice for this to be part of the page clone itself, i Imagine the issue there is that you need all the page ids to match the original default to the new - hence it comes after?

srh-sloan: Yeah that's exactly it - I can't update a page with its next default ID until that next page has been made, so much simpler to just do after.

tests/test_clone.py
srh-sloan: Thanks, was going for the second option so have refactored, hopefully clearer!

Adam-W1 (approved):
LGTM :)

PR #46, opened 2024-09-18
Bau fix deploys

srh-sloan:
### Change description
Fixing aws copilot deploys and tests:
- Setup to run the db migrations
- Marking use of `randint` as `#nosec` so it doesn't trigger a warning - this is used in temporary POC places and won't be used in production


NarenderRajuB (approved):
LGTM

PR #47, opened 2024-09-18
fs-4642 adding cascade deletes

srh-sloan:
### Change description
When editing an application round, the following now happens:
- If you delete a form from a section, the DB rows for that form, it's pages and components are also deleted
- If you delete a section from the application, then the forms within that section, and all their children are also deleted

### How to test
Create a round with multiple sections and forms within those sections
Delete a form and check it and its pages are deleted from the db
Delete a section and check it and it's forms are removed from the db

No UI changes as the behaviour was already present in the frontend but it was doing a soft delete so this change is backend only


NarenderRajuB (dismissed):
LGTM

NarenderRajuB (approved):
LGTM

PR #56, opened 2024-09-27
FS:4651:  Add Form to import new templates

NarenderRajuB:
FS:4651:  Add Form to import new templates

### Change description

### How to test
1. Click of View templates  will show below screen, Enter template name and select json file  and click to add the template.

image
image
image
image


app/blueprints/templates/routes.py
```diff
@@ -12,9 +18,32 @@
...
     sections = get_all_template_sections()
     forms = get_all_template_forms()
-    return render_template("view_templates.html", sections=sections, forms=forms)
+    form = TemplateUploadForm()
+    if form.validate_on_submit():
+        template_name = form.template_name.data
+        file = form.file.data
+        if not os.path.exists(file_upload_path):
+            os.makedirs(file_upload_path)
+        if file:
```
srh-sloan, line 40: Do we need to write the file to disk somewhere? Where will that be on AWS? Can't we just use the in-memory data from the upload and read the json? If we do need to write the file for some reason can we remove it after import please

srh-sloan (changes requested):
Looks good, pulled locally and working, just one query about the file

app/blueprints/templates/routes.py
NarenderRajuB: There are two options we can use
1) Save file in the container and deleter after the import /use flask-caching create a temp directory and then remove it.
2) If we need to refer to the file imported in future, then we may have to find a way to store in AWS.

Happy to change based on suggestions.

app/blueprints/fund_builder/routes.py
```diff
@@ -72,8 +76,15 @@ def section(round_id):
         "round_id": str(round_id),
     }
     existing_section = None
+    # TODO there must be a better way than a pile of ifs...
     if request.args.get("action") == "remove":
-        delete_section(section_id=request.args.get("section_id"), cascade=True)
+        delete_section_from_round(round_id=round_id, section_id=request.args.get("section_id"), cascade=True)
```
nuwan-samarasinghe, line 81: Shall we add some logs statements in to the critical points that will be easy for us to track down bugs also we'll add some comments

app/blueprints/templates/routes.py
nuwan-samarasinghe: As we discussed @NarenderRajuB  go ahead with the change & also check whether temp dir being deleted once done other wise these files will get piled up until a restart

tests/test_db.py
```diff
@@ -288,4 +298,293 @@ def test_form_sorting(seed_dynamic_data, _db):
...
     assert result_section.forms[2].form_id == formX.form_id
     assert result_section.forms[3].form_id == form2.form_id
-    assert result_section.forms[3].section_index == 3
+    assert result_section.forms[3].section_index == 4
+
+
+section_id = uuid4()
+
+
+@pytest.mark.seed_config(
```
nuwan-samarasinghe, line 307: Can we rename the file into `test_db_seed` or similar thing when we look the file name it seems like going to write db level tests

Adam-W1 (approved):
LGTM

app/blueprints/templates/routes.py
```diff
@@ -12,9 +19,32 @@
     template_folder="templates",
 )
 
+file_upload_path = os.path.join(os.path.dirname(__file__), "uplaoded_files")
```
nuwan-samarasinghe, line 22: DO we still need to upload & create a file inside the container cant we read the maltipartfile it self

nuwan-samarasinghe (changes requested):

app/blueprints/templates/routes.py
NarenderRajuB: Thanks @nuwan-samarasinghe  removed this, this was added earlier to save file, which is not used now

nuwan-samarasinghe (approved):
LGTM

PR #59, opened 2024-10-01
bau updating docker compose host names

srh-sloan:
### Change description
Updating the host names and container names in the dev docker compose file as they were conflicting with the dev container names


hamzabinkhalid (approved):
LGTM

PR #62, opened 2024-10-01
adding section_base_path column to round

srh-sloan:
Adding `section_base_path` column to Round table, with a sequence, and testing this


Adam-W1 (approved):
lgtm 👍

PR #75, opened 2024-10-17
Removing prod, tests that don't apply, and making name more accurate

srh-sloan:
Updates to the github actions file:
- Change run name to 'Build and Unit Test' instead of 'Dev Deploy' to match other apps, and because we aren't auto deploying to dev
- Commented out post deploy tests as they don't relate to FAB and hold things up if they fail as they are unrelated
- Commented out prod deploy and removed from options as we are a long way from prod so don't need it there atm


hamzabinkhalid (approved):
LGTM

PR #83, opened 2024-10-21
increasing length of short name and making description a multiline field

srh-sloan:
For COF EOI we need more than 6 characters in the short name field (as per existing config) so have increased the allowed length.
Also made description a multiline text field as it better fits the data


nuwan-samarasinghe (approved):
LGTM

hamzabinkhalid (approved):
LGTM

PR #84, opened 2024-10-22
Making contact details optional

srh-sloan:
Tiny tweak to add/edit round form:
- Contact textphone, telephone and email are now optional as COF doesn't have textphone or telephone
- Added 'en' to the fields that only need the English text at present


nuwan-samarasinghe (approved):
LGTM

PR #94, opened 2024-11-01
update readmes and config

Adam-W1:


nuwan-samarasinghe (approved):
LGTM

docs/run.md
```diff
@@ -0,0 +1,41 @@
...
+## Local development
+If developing locally, you will need a postgres instance running and to set the following environment variables:
+ - `DATABASE_URL` to a suitable connection string. The default used is
+ `postgresql://postgres:password@fab-db:5432/fund_builder`.   # pragma: allowlist secret
+ - `DATABASE_URL_UNIT_TEST` default
+ `postgresql://postgres:password@fab-db:5432/fund_builder_unit_test`  # pragma: allowlist secret
+
+## General
+Run the app with `flask run` (include `--debug` for auto reloading on file changes)
+
```
samuelhwilliams, line 14: How do these instructions intersect with the docker runner? Should we point to that as an alternative? Or should they be 'the way' to run all of our apps?

I suppose I wonder if theres now some many setup in each of our apps that talks about needing a DB, setting some vars etc that could all be removed in lieu of using the docker runner with a central set of config and instructions etc. But very possible I'm missing some developer need to run this or other apps individually?

Adam-W1: I think that's a great idea to simplify and centralise the READMEs, I suppose because FAB currently sits outside docker runner we have kept this separate for now. I agree though and to do this with live services repos might be the place to start.

PR #118, opened 2024-11-25
Add ctdf seed

srh-sloan:
Adds a tasks script to upsert the dummy fund (CTDF) from config


albertkol (approved):
:+1: It works!

samuelhwilliams:
@srh-sloan do we ... still need/want this?

PR #129, opened 2024-12-06
FS-4876 - Ensure all round fields are passed to the round export

wjrm500:
### Ticket

[Round fields not passed to export](https://mhclgdigital.atlassian.net/browse/FS-4876)

### Description

We noticed a bug whereby certain fields specified in the “Create a round” page of FAB are saved, but not passed to the export. This means that FAB-configured values for these values will have no effect in live services.

Specifically, the fields application_reminder_sent and eligibility_config are not exported.

### DIFF
--- a/app/export_config/generate_fund_round_config.py
+++ b/app/export_config/generate_fund_round_config.py
@@ -126,7 +126,7 @@ def generate_round_config(round_id):
         deadline=round.deadline.isoformat(),
         assessment_start=round.assessment_start.isoformat(),
         assessment_deadline=round.assessment_deadline.isoformat(),
-        application_reminder_sent=False,
+        application_reminder_sent=round.application_reminder_sent,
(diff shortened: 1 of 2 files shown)


app/export_config/generate_fund_round_config.py
```diff
@@ -126,7 +126,7 @@ def generate_round_config(round_id):
         deadline=round.deadline.isoformat(),
         assessment_start=round.assessment_start.isoformat(),
         assessment_deadline=round.assessment_deadline.isoformat(),
-        application_reminder_sent=False,
+        application_reminder_sent=round.application_reminder_sent,
```
nuwan-samarasinghe, line 129: Hope unit tests are testing these changes that we do

nuwan-samarasinghe (approved):
LGTM

PR #130, opened 2024-12-06
FS-4877 - Fix local testing of FAB when using global Docker runner

wjrm500:
### Ticket

[Fix local testing of FAB when using global Docker Runner](https://mhclgdigital.atlassian.net/browse/FS-4877)

### Description

Currently if we use the Funding Service-wide Docker runner to spin up apps locally, we get SQL connection issues when we try to run unit tests in FAB (uv run pytest .). This is because  these unit tests rely on the existence of a database fab_unit_test the global Docker Runner doesn’t provide this.

These changes are to rename from `fab_unit_test` to `fab_store_test` for consistency with the database name specified in [this PR](https://github.com/communitiesuk/funding-service-design-docker-runner/pull/100) and to change the environment variable name `DATABASE_URL_UNIT_TEST` to `DATABASE_URL_TEST` again for consistency with the database name and purpose.

### DIFF
--- a/.github/workflows/copilot_deploy.yml
+++ b/.github/workflows/copilot_deploy.yml
@@ -25,7 +25,7 @@ jobs:
     uses: communitiesuk/funding-service-design-workflows/.github/workflows/pre-deploy.yml@main
     with:
       postgres_unit_testing: true
-      db_name: fab_unit_test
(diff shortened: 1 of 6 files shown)


nuwan-samarasinghe (approved):
LGTM

PR #134, opened 2024-12-06
FS-4877 - Ensuring FAB unit tests run both locally and in GitHub

wjrm500:
### Ticket

[Fix local testing of FAB when using global Docker Runner](https://mhclgdigital.atlassian.net/browse/FS-4877)

### Description

This PR - https://github.com/communitiesuk/funding-service-design-fund-application-builder/pull/130 - was intended to fix this problem, but it did not. This is because of a discrepancy between the database password required for locally running unit tests versus running them in GitHub. I ran the tests locally with the password `password` which worked, then pushed the code up. The code failed in GitHub, so I changed the password to `postgres`, forgot to re-test and pushed the code up, which then passed on GitHub, then merged 🤓 

Anyway this PR fixes that by using an env var `DATABASE_URL` where available (containing the password `postgres`) and a hard-coded string containing the password `password` otherwise. This works because when running unit tests as part of the GitHub push workflow, we [inject DATABASE_URL]( https://github.com/communitiesuk/funding-service-design-workflows/blob/main/.github/workflows/pre-deploy.yml#L97-L100), providing a GitHub-compatible password, while when running locally, we have no env var set and so default to the locally-compatible connection string.

### DIFF
--- a/config/envs/unit_test.py
+++ b/config/envs/unit_test.py
@@ -16,7 +16,7 @@ class UnitTestConfig(Config):
     SECRET_KEY = "unit_test"  # pragma: allowlist secret
 
     SQLALCHEMY_DATABASE_URI = getenv(
-        "DATABASE_URL_TEST",
-        "postgresql://postgres:postgres@127.0.0.1:5432/fab_store_test",  # pragma: allowlist secret
+        "DATABASE_URL",
+        "postgresql://postgres:password@127.0.0.1:5432/fab_store_test",  # pragma: allowlist secret
     )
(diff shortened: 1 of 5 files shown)


samuelhwilliams (dismissed):
Tests work for me locally now 👍 

As per slack, some optional-but-nice extras:

* Can we get rid of other references to DATABASE_URL_TEST
* You might want to throw DATABASE_URL=postgresql://postgres:password@localhost:5432/fab_store_test  into pytest.ini and then add fab_store_test to the docker runner compose list of dbs to create

samuelhwilliams (approved):

PR #135, opened 2024-12-06
FSPT-119 Remove unused bundler dependencies

sfount:
We're not preprocessing any SCSS or bundling client side javascript so remove these dependencies, we can add them when they're needed.

As CSS is directly written and copied in this repository we can remove `pyscss`.


static_assets.py
```diff
@@ -19,28 +19,7 @@ def init_assets(app=None, auto_build=False, static_folder="app/static/dist"):
...
-        #     "./scss/*.scss",
-        #     filters="pyscss,cssmin",
-        #     output="css/main.min.css",
-        #     extra={"rel": "stylesheet/css"},
-        # )
-
-        # env.register("default_styles", css)
-        # env.register("main_js", js)
-
         bundles = []
```
samuelhwilliams, line 22: Does this file even ... do anything? There's no bundles, so nothing gets built? Or is there some magic here?

sfount: Hm yeh as far as I can tell it doesn't do anything - the only thing it does as an empty file is show how it could fit into the lifecycle of the app starting + a static `build.py` process. I think there's plenty of precedent to follow for that in other apps so we could just remove it here?

samuelhwilliams: yeah, I think if it's not used at all that sounds fair 👍 

Maybe one day it will be integrated... 👀

wjrm500 (dismissed):
Nice bit of cleaning up

PR #137, opened 2024-12-06
FS-4867 - Make feedback and guidance links non-mandatory in 'Create a Round' form

wjrm500:
### Ticket

[Make feedback and guidance links non-mandatory in "Create a Round"](https://mhclgdigital.atlassian.net/browse/FS-4867)

### Description

Feedback link and guidance URL are non-mandatory, therefore we should not mandate their collection in FAB.


NarenderRajuB (approved):
LGTM

PR #144, opened 2024-12-10
FS-4893 - Don't prepend 'Apply for' to exported summary page title

wjrm500:
### Ticket

[Summary Page Title](https://mhclgdigital.atlassian.net/browse/FS-4893)

### Description

Currently we prepend "Apply for" to the fund title as part of the FAB export. This causes weird wording on application summary page (Apply frontend) e.g., "Application for Apply for..."

### Testing instructions

_To follow..._

### DIFF
--- a/app/export_config/generate_form.py
+++ b/app/export_config/generate_form.py
@@ -284,7 +284,7 @@ def build_form_json(form: Form, fund_title: str = None) -> dict:
     """
 
     results = copy.deepcopy(BASIC_FORM_STRUCTURE)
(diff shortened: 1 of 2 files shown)


wjrm500:
On review, the issue here is that the fund title argument passed through from FAB is used in two different places in Apply:

- In the black banner at the top of the page, as “Apply for {fund_title}”
- In the title above the tasklist in the main body of the page, as “Application for {fund_title}”

This is therefore not a bug but rather a question of educating users on usage. The solution is to add help text which I believe is covered by another ticket, so I think we can discard the draft PR referenced earlier and close this ticket.

PR #146, opened 2024-12-11
FS-4899 - Create new list for each form even where list name (ID) is the same

wjrm500:
### Ticket

[Create new list for every form even where names collide](https://mhclgdigital.atlassian.net/browse/FS-4899)

### Description

Lists are scoped to forms. If two list-containing forms are loaded into FAB, even if those lists have the same name identifier, we should create two distinct lists in FAB, to preserve the data in those lists.

Currently, we do not do this. We check whether a list with the same name exists in the database, and if so we do not add the list.

This wouldn't be a practical problem if users created forms in isolation, as random identifiers are used for list names and these are unlikely to collide. The problem is that users clone forms, thereby cloning these random identifiers.

### DIFF
--- a/app/import_config/load_form_json.py
+++ b/app/import_config/load_form_json.py
@@ -7,7 +7,6 @@ from uuid import UUID
 
 from sqlalchemy.orm.attributes import flag_modified
 
-from app.db.queries.application import get_list_by_name
 from app.db.queries.application import insert_form_section
 from app.db.queries.application import insert_list
 from app.export_config.generate_form import human_to_kebab_case
@@ -128,19 +127,6 @@ def add_conditions_to_components(db, page: dict, conditions: dict, page_id):
                         component_to_update.conditions = [asdict(new_condition)]
 
 
-def _find_list_and_create_if_not_existing(list_name: str, all_lists_in_form: list[dict]) -> UUID:
-    list_from_form = next(li for li in all_lists_in_form if li["name"] == list_name)
-
-    # Check if this list already exists in the database
-    existing_list = get_list_by_name(list_name=list_name)
-    if existing_list:
-        return existing_list.list_id
-
-    # If it doesn't, insert new list
-    new_list = insert_list(do_commit=False, list_config={"is_template": True, **list_from_form})
-    return new_list.list_id
-
-
 def _find_form_section(form_section_name: str, form_section_list: list[dict]) -> UUID:
     form_section_to_search = form_section_name if form_section_name else "FabDefault"
     form_section_from_db = next(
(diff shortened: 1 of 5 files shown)


app/import_config/load_form_json.py
```diff
@@ -255,6 +238,11 @@ def insert_form_config(form_config, form_id):
     start_page_path = form_config["startPage"]
     form_section_list = create_form_sections_db(form_config)
 
+    list_names_to_ids = {}
+    for lizt in form_config.get("lists", []):
+        new_list = insert_list(do_commit=False, list_config={"is_template": True, **lizt})
```
nuwan-samarasinghe, line 243: in a Json there might have unwanted lists so don't we need to remove them from the template I think better to have list items which is already in use ? better to check first usage of the list in component then create ? 

also once template is deleted is the list also going to be deleted ?

tests/test_config_import.py
```diff
@@ -91,3 +97,49 @@ def test_import_multi_input_field(seed_dynamic_data, _db):
     assert multi_input_component
     assert multi_input_component.type == ComponentType.MULTI_INPUT_FIELD
     assert len(multi_input_component.children) == 4
+
+
+def test_creates_unique_lists_per_form_with_shared_list_name(app: Flask, clean_db: SQLAlchemy):
```
nuwan-samarasinghe, line 102: Can we have the delete functionality test as well ?

tests/conftest.py
```diff
@@ -43,6 +43,23 @@ def seed_dynamic_data(request, app, clear_test_data, _db, enable_preserve_test_d
     _db.session.commit()
 
 
+@pytest.fixture(scope="function")
+def clean_db(app, _db):
+    """Ensures a clean database before each test runs"""
```
nuwan-samarasinghe, line 48: can we verify is seed_dynamic_data also doing it or not ?

nuwan-samarasinghe (changes requested):

app/import_config/load_form_json.py
wjrm500: I think we mentioned this when we were pairing on these changes - we are trying to represent the Form Designer schema as much as possible in our own data schema - if Form Designer schema allows unattached lists, then so should we. The basic principle is minimising complexity our end

Regarding template deletion and list deletion, this is a good question - maybe we can create a separate ticket for ensuring cascade delete across the data schema as an outcome from your excellent work on the [FAB code end-to-end understanding](https://mhclgdigital.atlassian.net/browse/FS-4900) ticket?

tests/test_config_import.py
wjrm500: Delete functionality hasn't been changed here and I reckon we probably want to scope that separately; as per comment above:

> Regarding template deletion and list deletion, this is a good question - maybe we can create a separate ticket for ensuring cascade delete across the data schema as an outcome from your excellent work on the [FAB code end-to-end understanding](https://mhclgdigital.atlassian.net/browse/FS-4900) ticket?

Assuming we do this, we can add tests as part of this work to ensure cascade delete works properly. Good thinking to test any new functionality though 👍

tests/conftest.py
wjrm500: This was verified as part of my work on the ticket - `seed_dynamic_data` does not clear the database beforehand which means our test would be unstable using this, because we are checking that there are exactly two `Lizt` items in the database

nuwan-samarasinghe (approved):
LGTM

PR #149, opened 2024-12-13
FS-4895: When creating a fund or a round check, whether given round shortnames are available

nuwan-samarasinghe:
**Ticket: https://mhclgdigital.atlassian.net/browse/FS-4895**

### Change description
In this change we are focussing on fund and round short name reuse issue so following scenarios are acceptable 

1. If there is a fund and that given short name is not in our records that short name is acceptable 
2. If there is an update for the given fund short name is acceptable because it belongs to it self 
3. IF there is a fund and when round creation if the short name is not available then acceptable 
4. IF there is a different fund and we are planing to use same short name it is acceptable 
5. If we do update the round then it will be again acceptable since short name belongs to it self

Following are not acceptable 

1. If there is a fund for a given short name then we are shoving an error for validation purpose 
2. If there is a fund and inside that fund if we have same short name round we give an error 

### How to test

**Testing Fund create**

1. Use FAB 
2. Try to create a fund that does not exists then it should show the success 
3. Try to create a existing short named fund it gives error
4. Try to rename a short name of a fund to a already existing one then it should give an error other wise should be success

**Testing Round Create**

1. Use FAB
2. Try to rename a short name of a round to a already existing one then it should give an error other wise should be success

image

image

### DIFF
--- a/app/blueprints/fund_builder/routes.py
+++ b/app/blueprints/fund_builder/routes.py
@@ -254,49 +254,41 @@ def fund(fund_id=None):
     else:
         form = FundForm()
 
-    error = {}
     if form.validate_on_submit():
-        is_short_name_available = False
-        if form.data and form.data['short_name']:
-            fund_data = get_fund_by_short_name(form.data['short_name'])
-            is_short_name_available = fund_data and str(fund_data.fund_id) != fund_id
-        if not is_short_name_available:
-            if fund_id:
-                fund.name_json["en"] = form.name_en.data
-                fund.name_json["cy"] = form.name_cy.data
-                fund.title_json["en"] = form.title_en.data
-                fund.title_json["cy"] = form.title_cy.data
-                fund.description_json["en"] = form.description_en.data
-                fund.description_json["cy"] = form.description_cy.data
-                fund.welsh_available = form.welsh_available.data == "true"
-                fund.short_name = form.short_name.data
-                fund.audit_info = {"user": "dummy_user", "timestamp": datetime.now().isoformat(), "action": "update"}
-                fund.funding_type = form.funding_type.data
-                fund.ggis_scheme_reference_number = (
-                    form.ggis_scheme_reference_number.data if form.ggis_scheme_reference_number.data else ""
-                )
-                update_fund(fund)
-                flash(f"Updated fund {form.title_en.data}")
-                return redirect(url_for("build_fund_bp.view_fund", fund_id=fund.fund_id))
-
-            new_fund = Fund(
-                name_json={"en": form.name_en.data},
-                title_json={"en": form.title_en.data},
-                description_json={"en": form.description_en.data},
-                welsh_available=form.welsh_available.data == "true",
-                short_name=form.short_name.data,
-                audit_info={"user": "dummy_user", "timestamp": datetime.now().isoformat(), "action": "create"},
-                funding_type=FundingType(form.funding_type.data),
-                ggis_scheme_reference_number=(
-                    form.ggis_scheme_reference_number.data if form.ggis_scheme_reference_number.data else ""
-                ),
+        if fund_id:
+            fund.name_json["en"] = form.name_en.data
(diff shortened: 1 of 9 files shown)


app/blueprints/fund_builder/routes.py
```diff
@@ -254,42 +254,50 @@ def fund(fund_id=None):
...
-        add_fund(new_fund)
-        flash(f"Created fund {form.name_en.data}")
-        return redirect(url_for(BUILD_FUND_BP_INDEX))
+            add_fund(new_fund)
+            flash(f"Created fund {form.name_en.data}")
+            return redirect(url_for(BUILD_FUND_BP_INDEX))
 
-    error = error_formatter(form.errors)
+        error = {**error,
+                 'short_name': [f'Given fund short name already exists.']}
```
samuelhwilliams, line 299: this doesn't seem familiar to me as GOV.UK error message content style - do we want to run this through content design? Or is that all happening separately?

wjrm500 (changes requested):
Hey @nuwan-samarasinghe, I think a better implementation would be to create validation functions and attach them directly to the FlaskForms e.g.,

In `apps/funding-service-design-fund-application-builder/app/blueprints/fund_builder/forms/fund.py`:

```
...
def unique_fund_short_name(form, field):
    fund = get_fund_by_short_name(field.data)
    if fund and str(fund.fund_id) != form.fund_id.data:
        raise ValidationError("Given fund short name already exists.")
...
class FundForm(FlaskForm):
    ...
    short_name = StringField(
        "Short name",
        validators=[
            DataRequired(),
            Length(max=10),
            no_spaces_between_letters,
            unique_fund_short_name,
        ],
    )
    ...
```

This way we don't need to modify the route function or the error formatter at all, and we achieve the same result.

We can then do the same for Round. Let me know what you reckon!

app/blueprints/fund_builder/routes.py
nuwan-samarasinghe: We have a ticket to verify the error message with content designer @chaneldiep

NarenderRajuB (approved):
LGTM

NarenderRajuB (dismissed):
LGTM

tests/test_routes.py
```diff
@@ -115,6 +147,66 @@ def test_update_fund(flask_test_client, seed_dynamic_data):
...
+        "feedback_link": "http://example.com/feedback",
+        "project_name_field_id": 1,
+        "guidance_url": "http://example.com/guidance",
+    }
+
+    response = submit_form(flask_test_client, "/round", new_round_data)
+    assert response.status_code == 200
+    new_round_data = {**new_round_data, 'short_name': 'NR1234'}
+    response = submit_form(flask_test_client, "/round", new_round_data)
+    assert response.status_code == 200
```
wjrm500, line 203: Could we have assertions to check the error message doesn't appear in the HTML for the first two submissions? As we expect a 200 response either way. We could have:

```
error_html = '<a href="#short_name">Short name: Given short name already exists in the fund funding to improve testing.</a>'

# Test works fine with first round
response = submit_form(flask_test_client, "/round", new_round_data)
assert response.status_code == 200
assert error_html not in response.data.decode('utf-8'), "Error HTML found in response"

# Test works fine with second round but with different short name
new_round_data = {**new_round_data, 'short_name': 'NR1234'}
response = submit_form(flask_test_client, "/round", new_round_data)
assert response.status_code == 200
assert error_html not in response.data.decode('utf-8'), "Error HTML found in response"

# Test doesn't work with third round with same short name as firsrt
new_round_data = {**new_round_data, 'short_name': 'NR123'}
response = submit_form(flask_test_client, "/round", new_round_data)
assert response.status_code == 200
assert error_html in html = response.data.decode('utf-8'), "Error HTML not found in response"
```

wjrm500 (changes requested):
Just a couple of small changes and we're good to go!

wjrm500 (approved):

PR #150, opened 2024-12-13
FS-4833 - Adding basic auth to FAB

wjrm500:
### Ticket

[Add basic auth to FAB for pre-prod environments](https://mhclgdigital.atlassian.net/browse/FS-4833)

### Description

Following the convention of other apps including [Authenticator](https://github.com/communitiesuk/funding-service-design-authenticator/pull/358/files), we can use an Nginx sidecar to seamlessly provide basic auth on all HTTP requests, while allowing requests to `/healthcheck` through.

We have only added to Dev and Test environments, not UAT as this will be our _de facto_ prod environment for the January delivery, and will be authenticated by Entra ID.

Note that we also need to ensure values are in the AWS Parameter Store to support this change.

### DIFF
--- a/copilot/fsd-fund-application-builder/manifest.yml
+++ b/copilot/fsd-fund-application-builder/manifest.yml
@@ -64,9 +64,41 @@ environments:
   dev:
     count:
       spot: 1
+    sidecars:
+      nginx:
+        port: 8087
+        image:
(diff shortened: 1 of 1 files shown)


hamzabinkhalid (approved):
Looks good 👍

PR #158, opened 2024-12-17
FS-4848 - Add field names to error messages in error message banner

wjrm500:
### Ticket

[Add an error banner to notify users when there are issues with mandatory fields (Add a Round)](https://mhclgdigital.atlassian.net/browse/FS-4848)

### Description

Add field names to error messages in error message banner.


nuwan-samarasinghe (approved):
LGTM

PR #160, opened 2024-12-17
BAU - Add page title of 'Fund Application Builder' by default

wjrm500:
Currently we have no page title at all as neither `service_title` nor `pageHeading` are set anywhere. The code is effectively broken. This is just a very quick and hopefully uncontroversial fix to set "Fund Application Builder" as the page title in the base template.


NarenderRajuB (approved):
LGTM

nuwan-samarasinghe (approved):

PR #161, opened 2024-12-17
Create assessment config

albertkol:
Create assessment config


albertkol:
I don't think this will be merged, but perhaps the internet will find some use in the wisdom behind my code. o7

PR #173, opened 2024-12-23
FS-4922 - Handle 500 error gracefully with error page render

wjrm500:
### Ticket

https://mhclgdigital.atlassian.net/browse/FS-4922

### Description

Adds default error handler to application to catch all internal server errors and render a new 500.html page, which looks like this:

image


nuwan-samarasinghe (approved):
LGTM

PR #180, opened 2025-01-02
FS-4924 - Navigation bar

wjrm500:
### Ticket

https://mhclgdigital.atlassian.net/browse/FS-4924

### Description

- GOV.UK Frontend updated from 5.4.0 to 5.6.0 to include new ["Service navigation" component](https://design-system.service.gov.uk/components/service-navigation/)
- `navigation.html` partial added to be conditionally rendered by `base.html` according to variable `showNavigationBar`, which can be set further up the hierarchy e.g., `dashboard.html`
- `current` status of nav items controlled by `active_item_identifier` variable, also set further up the hierarchy e.g., `dashboard.html`

### Screenshot of UI changes

image


app/blueprints/index/templates/dashboard.html
```diff
@@ -1,4 +1,7 @@
 {% extends "base.html" %}
+{% set showNavigationBar = True %}
+{% set active_item_identifier = "home" %}
```
nuwan-samarasinghe, line 3: Cant we send this variable from the endpoint side like the following 
`return render_template("home.html", showNavigationBar=True, active_item_identifier="home")`

instead having as a variable in nunjux side because we should focus on presentation side not logical side in templating

app/templates/base.html
```diff
@@ -10,20 +10,27 @@
 
 {% block header %}
   {{ govukHeader({
-    'useTudorCrown': 'yes',
-    'serviceName': 'Fund Application Builder',
-    'serviceUrl': url_for('index_bp.index'),
+    "useTudorCrown": "yes",
+    "serviceName": "Fund Application Builder",
```
nuwan-samarasinghe, line 14: Cant we send these from the python side may be as a meta data instead of having hard coded data in the templates ?

nuwan-samarasinghe (changes requested):

app/blueprints/index/templates/dashboard.html
wjrm500: Hi @nuwan-samarasinghe - yes technically we could do that, but the variable values are intrinsically linked to the template and so it doesn't make sense to me that you would be able to render the template with different values. For example, why would you want to be able to render the home page without the navigation bar, or with a page that isn't the home page set as active?

app/templates/base.html
wjrm500: Hey - yes we could, but not sure why we'd want to? This is a base template used by all pages and although we could in theory set the service name as an environment variable for example, there doesn't seem to be huge benefit. Regardless of what's best though, the lines changed in the `govUkHeader` params are just quote -> double quote so we shouldn't block merge here

nuwan-samarasinghe (approved):

PR #186, opened 2025-01-06
FS-4945 - Tidy dependencies

wjrm500:
### Ticket

https://mhclgdigital.atlassian.net/browse/FS-4945

### Changes made in this PR

- **Removing unused non-dev dependencies** - these dependencies are not used by the application and should be removed
- **Moving beautifulsoup4 dependency from dev dependency group to main project dependencies** - beautifulsoup4 is required for the application to function - it is imported in `app/all_questions/metadata_utils.py`, which is imported in `app/blueprints/application/routes.py` among other places

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,24 +8,16 @@ license = "MIT License"
 requires-python = "~=3.10.0"
 dependencies = [
     "airium==0.2.6",
-    "flask-babel==4.0.0",
-    "flask-compress==1.15",
     "flask-migrate==4.0.7",
     "flask-sqlalchemy==3.1.1",
-    "flask-talisman==1.1.0",
     "flask-wtf==1.2.2",
     "flask==3.0.3",
     "funding-service-design-utils==5.2.0",
     "govuk-frontend-jinja==3.4.0",
     "jsonschema==4.23.0",
-    "marshmallow-sqlalchemy==1.0.0",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -8,25 +8,17 @@ license = "MIT License"
...
     "jsonschema==4.23.0",
-    "marshmallow-sqlalchemy==1.0.0",
-    "openpyxl==3.1.5",
-    "pandas==2.2.3",
     "psycopg2-binary==2.9.10",
     "requests==2.32.3",
     "sentry-sdk==2.10.0",
-    "sqlalchemy-json==0.7.0",
-    "sqlalchemy-utils==0.41.2",
-    "sqlalchemy[mypy]==2.0.36",
```
samuelhwilliams, line 29: I would say that this is still a core dependency, even if we're also using `flask-sqlalchemy` which brings it in. It's a very fundamental part of our application (managing models+DB stuff), so I think maybe we keep this one?

samuelhwilliams (approved):
Looks very sensible - one comment - otherwise nice ✅ 

We probably want to implement flask-talisman or similar at some point because it manages a bunch of security configuration/enforcement, eg HTTPS, CSP, etc. Fine to remove now if we're not using it, but probably should be a backlog ticket somewhere to add it back and set some sensible settings.

PR #187, opened 2025-01-07
FS-4947 - Implement Flask-Talisman security headers

wjrm500:
### Ticket
[FS-4947](https://mhclgdigital.atlassian.net/browse/FS-4947)

### Changes made

- **Remove unused Sortable.js** - Removed an unused script that was being loaded from an external CDN. We don't want to include it as a safe script source in our Flask-Talisman Content Security Policy (CSP)
- **Move inline JavaScript to files** - Replaced inline JavaScript with external file references in the templates for creating and editing Fund and Round and modified the static asset build process to handle this. When using Flask-Talisman, inline JavaScript necessitates adding script hashes to the CSP, which is not maintainable
- **Add Flask-Talisman** - Implemented Flask-Talisman with settings copied from [Pre-Award](https://github.com/communitiesuk/funding-service-pre-award/blob/main/config/envs/default.py). Replaced Pre-Award sources in the CSP with ones specifically relevant to FAB

### Testing instructions

1. Content Security Policy

- Visit `/funds/create` or access an existing fund
- Toggle the "Welsh available" radio buttons
- Verify the Welsh fields appear/disappear smoothly with no console errors

This confirms our external JavaScript is correctly whitelisted in CSP while inline scripts are properly removed

2. Basic Security Headers

- Using browser developer tools, inspect response headers for any page
- Verify presence of:
  - Content-Security-Policy with default-src 'self'
  - X-Frame-Options: SAMEORIGIN
  - Referrer-Policy: strict-origin-when-cross-origin

This confirms Flask-Talisman is active with our configuration

### DIFF
--- a/config/envs/default.py
+++ b/config/envs/default.py
@@ -34,11 +34,7 @@ class DefaultConfig(object):
     # Content Security Policy
     SECURE_CSP = {
         "default-src": "'self'",
-        "script-src": [
-            "'self'",
-            "'sha256-GUQ5ad8JK5KmEWmROf3LZd9ge94daqNvd8xy9YS1iDw='",
-            "'sha256-0F02u9outG+s4SWeWHc/Pmk8iJNh7GtqrTtLPzkxZ3M='",
-        ],
+        "script-src": ["'self'"],  # Needed to enable nonce
     }
 
     # Talisman Config
@@ -75,7 +71,6 @@ class DefaultConfig(object):
         "content_security_policy": SECURE_CSP,
         "content_security_policy_report_uri": None,
         "content_security_policy_report_only": False,
-        # "content_security_policy_nonce_in": None,
         "referrer_policy": FSD_REFERRER_POLICY,
         "session_cookie_secure": True,
         "session_cookie_http_only": True,
--- a/config/envs/default.py
+++ b/config/envs/default.py
@@ -30,3 +30,57 @@ class DefaultConfig(object):
     RSA256_PUBLIC_KEY_BASE64 = getenv("RSA256_PUBLIC_KEY_BASE64")
     if RSA256_PUBLIC_KEY_BASE64:
         RSA256_PUBLIC_KEY: str = base64.b64decode(RSA256_PUBLIC_KEY_BASE64).decode()
+
+    # Content Security Policy
+    SECURE_CSP = {
+        "default-src": "'self'",
+        "script-src": [
+            "'self'",
+            "'sha256-GUQ5ad8JK5KmEWmROf3LZd9ge94daqNvd8xy9YS1iDw='",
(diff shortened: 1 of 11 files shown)


config/envs/default.py
```diff
@@ -30,3 +30,57 @@ class DefaultConfig(object):
...
     if RSA256_PUBLIC_KEY_BASE64:
         RSA256_PUBLIC_KEY: str = base64.b64decode(RSA256_PUBLIC_KEY_BASE64).decode()
+
+    # Content Security Policy
+    SECURE_CSP = {
+        "default-src": "'self'",
+        "script-src": [
+            "'self'",
+            "'sha256-GUQ5ad8JK5KmEWmROf3LZd9ge94daqNvd8xy9YS1iDw='",
+            "'sha256-0F02u9outG+s4SWeWHc/Pmk8iJNh7GtqrTtLPzkxZ3M='",
```
samuelhwilliams, line 40: This is something that hasn't been done well in pre-award, but that I'd recommend: can you add a comment next to each line indicating what script this is allowing? Makes it a bit more auditable and understandable 🙏  I'm assuming it's the two scripts you've converted from `<script>...</script>` to `<script src=""/>`

samuelhwilliams: If it _is_ this files, consider using CSP nonces instead of these script hashes - should be a bit more maintainable as the hashes won't need updating if the file contents change:

https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/a01b416375eef42cb70bb7dd13b856581f178dce/app.py#L158

and 

image

config/envs/default.py
```diff
@@ -30,3 +30,57 @@ class DefaultConfig(object):
     RSA256_PUBLIC_KEY_BASE64 = getenv("RSA256_PUBLIC_KEY_BASE64")
     if RSA256_PUBLIC_KEY_BASE64:
         RSA256_PUBLIC_KEY: str = base64.b64decode(RSA256_PUBLIC_KEY_BASE64).decode()
+
+    # Content Security Policy
+    SECURE_CSP = {
+        "default-src": "'self'",
+        "script-src": [
+            "'self'",
+            "'sha256-GUQ5ad8JK5KmEWmROf3LZd9ge94daqNvd8xy9YS1iDw='",
```
nuwan-samarasinghe, line 39: For the CSS files that is going to render the app what will be the case for those ? for CSS don't we need to worry ? and also do we have any other JS coming from the govuk dependencies ?

nuwan-samarasinghe (approved):
Added some comments but overall it is good and saw there are few comments from @samuelhwilliams please check those

wjrm500:
@samuelhwilliams @nuwan-samarasinghe - have implemented with nonces instead of script hashes, see latest commit

samuelhwilliams (approved):
This looks good to me. Confusingly some of the session cookie settings don't work if the app is running locally in debug mode, so got stuck investigating that for a bit. But I think we're good 👍

PR #193, opened 2025-01-09
FS-4925 - Flow between "Create" pages

wjrm500:
### Ticket

https://mhclgdigital.atlassian.net/browse/FS-4925

### Understanding the changes

- **Splitting fund route function into create_fund and edit_fund** - Split the fund route into separate create and edit functions to make flow clearer, in light of upcoming addition of complexity including conditional redirect
- **Adding new save flow to create / edit fund page** - Updated the create / edit fund template with "Save and continue", "Cancel" and "Save and return to home" actions. "Cancel" takes users to the dashboard without saving data; "Save and return to home" redirects users to the dashboard if form validation succeeds, using a new `save_dest` query parameter. "Save and continue" redirects users if form validation succeeds to the "Create a Round" page if creating a fund, or to the "Fund config" page if updating a fund
- **Adding new save flow to create / edit application page** - Applying the above logic but to applications instead of funds. "Save and continue" redirects users if form validation succeeds to the "Build Application" page if creating an application, or to the "Fund config" page if updating an application
- **Secure URL redirection validation for save_dest query parameter** - Implements a secure URL validation pattern to prevent open redirect vulnerabilities in fund and application creation and editing endpoints
- **No longer depending on JavaScript for 'Save and return to home' functionality** - Moved from using JavaScript and a URL query parameter for handling the "Save and return home" functionality to using a standard HTML form submission with an "action" button value that determines the redirect destination.

### Testing the changes

**Pre-requisites**

- Checkout this branch locally
- Spin up FAB container (use [global Docker runner](https://github.com/communitiesuk/funding-service-design-docker-runner))

**Test suite 1 - Create a Fund**
- Action: Click "Add a new grant" from dashboard, fill in required fields, click "Save and continue"
- Expected behaviour: Form saves and you are redirected to "Create a Round"
- Action: Click "Cancel" from "Create a Fund" page
- Expected behaviour: Returns to dashboard without saving changes
- Action: Click "Save and return to home" from "Create a Fund" page
- Expected behaviour: Form saves and redirects to dashboard

**Test suite 2 - Create a Round**
- Action: Click "Create a new application", select a fund, fill required fields, click "Save and continue"
- Expected behaviour: Form saves and you are redirected to "Build Application"
- Action: Click "Cancel" from "Create a Round" page
- Expected behaviour: Returns to dashboard without saving changes
- Action: Click "Save and return to home" from "Create a Round" page
- Expected behaviour: Form saves and redirects to dashboard
- 
**Test suite 3 - Edit a Fund**
- Access: Navigate to `/funds/view`, select a fund and click "Edit Fund" (underneath the "Fund Meta Data" table)
- Action: Modify fields, click "Save and continue"
- Expected behaviour: Form saves and you are redirected back to "Fund Settings"
- Action: Click "Cancel" from "Edit Fund" page
- Expected behaviour: Returns to dashboard without saving changes
- Action: Click "Save and return to home" from "Edit Fund" page
- Expected behaviour: Form saves and redirects to dashboard

**Test suite 4 - Edit a Round**
- Access: Navigate to `/funds/view`, select a fund that you know has at least one round / application, expand one of the rounds underneath the "Application Rounds" heading, and click "Edit Round" (underneath the table)
- Action: Modify fields, click "Save and continue"
- Expected behaviour: Form saves and you are redirected back to "Fund Settings"
- Action: Click "Cancel" from "Edit Round" page
- Expected behaviour: Returns to dashboard without saving changes
- Action: Click "Save and return to home" from "Edit Round" page
- Expected behaviour: Form saves and redirects to dashboard

### UI screenshot

image


app/shared/helpers.py
```diff
@@ -71,3 +71,13 @@ def all_funds_as_govuk_select_items(all_funds: list) -> list:
     in the html
     """
     return [{"text": f"{f.short_name} - {f.name_json['en']}", "value": str(f.fund_id)} for f in all_funds]
+
+
+def is_safe_url(url):
+    """
+    Check if a URL is safe to redirect to. Only allows alphanumeric chars, forward slashes, hyphens and underscores.
+    """
+    if not url:
```
nuwan-samarasinghe, line 80: I just check is there any implementations to check the safe URl so found these 

[Flask Documentation](https://flask-login.readthedocs.io/en/latest/#login-example)
[Blog](https://stackoverflow.com/questions/60532973/how-do-i-get-a-is-safe-url-function-to-use-with-flask-and-how-does-it-work)

app/static/src/js/create_edit_fund.js
```diff
@@ -22,3 +22,14 @@ document.addEventListener('DOMContentLoaded', function() {
         radio.addEventListener('change', toggleWelshFields);
     });
 });
+
+// Add click event listener to save and return to home link
+document.getElementById('save-return-home').addEventListener('click', function(e) {
```
nuwan-samarasinghe, line 27: Check is event listeners are staking up if those stack up browser slowness can be seen and some anomalies

app/static/src/js/create_edit_round.js
```diff
@@ -17,3 +17,14 @@ function toggleWelshFields() {
 document.addEventListener('DOMContentLoaded', function() {
     toggleWelshFields();
 });
+
+// Add click event listener to save and return to home link
+document.getElementById('save-return-home').addEventListener('click', function(e) {
```
nuwan-samarasinghe, line 22: Check is event listeners are staking up if those stack up browser slowness can be seen and some anomalies

nuwan-samarasinghe (changes requested):

app/static/src/js/create_edit_fund.js
```diff
@@ -22,3 +22,14 @@ document.addEventListener('DOMContentLoaded', function() {
...
+    e.preventDefault();
```
nuwan-samarasinghe, line 28: Also cant we do this in the python side ?

app/blueprints/fund/routes.py
```diff
@@ -103,7 +63,63 @@ def fund(fund_id=None):
...
+    error = error_formatter(form)
+    return render_template("fund.html", form=form, fund_id=None, error=error)
+
+
+@fund_bp.route("/<fund_id>", methods=["GET", "POST"])
+def edit_fund(fund_id):
+    """Updates an existing fund"""
+    fund = get_fund_by_id(fund_id)
+
+    if request.method == "GET":
```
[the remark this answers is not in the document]
samuelhwilliams: Just dropping in - platform team are currently exploring rewriting a vertical slice of part of the pre-award service with the aim of making some of these decisions.

I might suggest in the short-term conforming to whatever your existing standard is until that work is done, and then we can start consistently applying the new standard (which is almost certainly MVC-like) everywhere?

But also - feel free to ignore me if you think this is a good improvement to make now - just sharing that maybe we're both thinking about the same sort of things right now.

app/static/src/js/create_edit_fund.js
samuelhwilliams: 100% I agree - we should do this with standard HTML POST; I don't think we want JS here.

It would be nice if we only use JS for progressive enhancement, not basic functionality 🙏

app/blueprints/fund/templates/fund.html
NarenderRajuB, line 1: {% set pageHeading %}{{ 'Update a Fund' if fund_id else 'Create a Fund' }} {% endset %}

i guess we need to change here as per figma
{% set pageHeading %}{{ 'Update a Grant' if fund_id else 'Create a Grant' }} {% endset %}

app/static/src/js/create_edit_fund.js
wjrm500: Noted, will look at removing the JS and improving the logic now

app/blueprints/fund/routes.py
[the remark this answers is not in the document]
wjrm500: @nuwan-samarasinghe although we have these nominal `services.py` modules in FAB, we don't have a strong concept of a service layer right now so happy to stick with existing pattern, at least on this PR, with a view to potentially re-architecting in future in line with Platform's vision for software architecture. Great to be thinking about this sort of thing though 🙏

app/static/src/js/create_edit_fund.js
wjrm500: Modified now 👍

app/static/src/js/create_edit_fund.js
wjrm500: Resolving as have removed this JS

app/shared/helpers.py
wjrm500: Resolving as have removed this function because we have also removed `save_dest` param

app/blueprints/fund/templates/fund.html
wjrm500: Good spot but for a separate ticket / PR I think as not in scope of this one

nuwan-samarasinghe (approved):
LGTM

PR #197, opened 2025-01-10
FS-4929: Adding all grants view page

nuwan-samarasinghe:
**Ticket : https://mhclgdigital.atlassian.net/browse/FS-4929**

### Change description
In this PR we have created the View All grants view and here we are using the generic table page component

Following is the proposed figma design 

image

Components Available
1. Title
2. Title description
3. Detail
4. Table with 3 columns Grant, Description, Grant type and Grant will be a clickable link
5. Add new grant (new grant creation)
6. Pagination is addressed in following [PR](https://github.com/communitiesuk/funding-service-design-fund-application-builder/pull/192)

Following features are available 

1. Add new grant will direct the user to create a new grant 
7. Grant name clickable link will open the detail view of the grant details (@NarenderRajuB  is working on that)

### How to test
[Link for the grants](https://fund-application-builder.levellingup.gov.localhost:3011/grants/)

image


wjrm500 (approved):
Looking beautiful, great to see FAB coming along so well with all these changes over the past couple of days. There are some minor issues but nothing to block merge. For the record though, here are my thoughts on said minor issues...

- Current way in which we pass GenericTablePage object to template in params is a bit weird, we basically turn it into a dict of itself to give `{"generic_table_page": generic_table_page}` leading to confusing code like `params = GenericTablePage(...)`, also we end up with `generic_table_page` being referred to by that name in the grants table template which also feels wrong (as it's not generic at that point)
- Passing HTML string to flash function is functional but mixes concerns, with presentation logic inside Python code. Need to think of a better approach
- Likewise `build_fund_rows` has some HTML in
- The new test seems slightly like overkill with all of the content checks, means any content changes will need to be made in two places

Great work though ❤️

PR #199, opened 2025-01-10
FS:4590- Implement Signout for FAB

NarenderRajuB:
Ticket : https://mhclgdigital.atlassian.net/browse/FS-4950

### Change description
Add Sign out functionality for FAB.
1. Internal users authenticated can now signout using Sign out link available in header.
2. Sign out is available in all pages with in FAB from Header only After user is logged in

- [x ] Unit tests and other appropriate tests added or updated
- [x ] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### How to test
1. Use FAB
2. Login to FAB 
3. Sign out link is available for the user to logout
4. Once the user clicks on  Sign out , user is logged out and  redirect to Authenticator login page, on login it redirects to FAB Dashboard/Home page

image
image

### DIFF
--- a/app/templates/base.html
+++ b/app/templates/base.html
@@ -15,7 +15,12 @@
     "useTudorCrown": "yes",
     "serviceName": "Fund Application Builder",
     "serviceUrl": url_for("index_bp.index"),
-    "classes": "govuk-header--full-width-border"
+    "navigation": [
+        {
+            "text": "Sign out",
+            "href": g.logout_url
+        }
+    ]  if g.is_authenticated else []
   }) }}
 
   {# Only show the custom navigation if showNavigationBar is set #}
--- a/config/envs/default.py
+++ b/config/envs/default.py
@@ -14,7 +14,9 @@ class DefaultConfig(object):
     FLASK_ENV = CommonConfig.FLASK_ENV
     SECRET_KEY = CommonConfig.SECRET_KEY
 
-    FAB_HOST = getenv("FAB_HOST", "fab:8080/")
+    FUND_APPLICATION_BUILDER_HOST = getenv(
+        "FUND_APPLICATION_BUILDER_HOST", "https://fund-application-builder.levellingup.gov.localhost:3011"
+    )
     FAB_SAVE_PER_PAGE = getenv("FAB_SAVE_PER_PAGE", "dev/save")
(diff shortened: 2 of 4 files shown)


wjrm500 (changes requested):
Hey @NarenderRajuB, couple of major issues:

- I can see from testing that the logout functionality does not actually work - did you test this yourself? Clearing the session like you are doing doesn't work, because we use a JWT cookie. Further, we need to actually terminate the session with identity provider i.e., Entra ID, which we can only do via Authenticator. We need to use Authenticator's sign out endpoint `/sessions/sign-out`
- The button is not in the right place, according to Mika's design. It should be to the right of the page, not beneath the title

NarenderRajuB:
> * /sessions/sign-out

@wjrm500 we can use g.logout_url , but for this , user will not be redirected to  FAB login screen , eventhough the Authenticator login button will be available to login user back.  Regarding the alignment after checking with @mikaelmetthey  i have added in that place. GovUKHeader() uses the Navigation attribute to be rendered in next line. We can look into this

app/templates/base.html
```diff
@@ -15,7 +15,14 @@
     "useTudorCrown": "yes",
     "serviceName": "Fund Application Builder",
     "serviceUrl": url_for("index_bp.index"),
-    "classes": "govuk-header--full-width-border"
+    "classes": "govuk-header--full-width-border",
```
nuwan-samarasinghe, line 18: I think for now since we don't have css override we are going with the default way to show links of logout in the FAB ?

nuwan-samarasinghe: Have you tested manually that once logged out is it deleting the user session cookie ?

nuwan-samarasinghe (changes requested):

app/templates/base.html
NarenderRajuB: @nuwan-samarasinghe yes session cookie is deleted when we click signout which is session/signout from authenticator

gidsg (approved):

nuwan-samarasinghe (approved):
LGTM

wjrm500:
@mikaelmetthey just double checking that you have signed off on this design, as it is different to the Figma? See screenshot below:

image

PR #200, opened 2025-01-10
Bugfix - Fixes grants table pagination

wjrm500:
We need to pass current page to GenericTablePage in the view_all_funds route in order for pagination to work.


wjrm500:
Merged without review as this is a clear and obvious bugfix for the [PR](https://github.com/communitiesuk/funding-service-design-fund-application-builder/pull/197) myself and @nuwan-samarasinghe merged less than an hour ago

PR #203, opened 2025-01-13
FS-4935: Grant details template and API 

NarenderRajuB:
Ticket: https://mhclgdigital.atlassian.net/browse/FS-4935

### Change description
In this Ticket grant_details API is added to render the grant details when clicked on grant name in Grants list page.

- [x ] Unit tests and other appropriate tests added or updated
- [ x] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### How to test
1. Use FAB
2. Create a fund/grant
3. Click on Grants menu to list all grants.
4. Click on one of the Grant name , This will render a grant detail page.
ex: https://fund-application-builder.levellingup.gov.localhost:3011/grants/details?grant_id=f9aa36e3-d42e-4416-9113-bd47dbd0a6f3
5. Clicking on Back button should redirect to Grants list page.
6. Clicking on change link for the field will redirect to  edit grant/fund page with focus on the field selected for change
7. Clicking on "Continue and Save" will redirect to grant details page

image
image
image

### DIFF
--- a/app/blueprints/fund/routes.py
+++ b/app/blueprints/fund/routes.py
@@ -64,6 +64,16 @@ def view_fund():
     return render_template("fund_config.html", **params)
 
 
+@fund_bp.route("/<uuid:fund_id>", methods=["GET"])
+def view_fund_details(fund_id):
+    """
+    Renders grant details page
+    """
+    form = FundForm()
+    fund = get_fund_by_id(fund_id)
+    return render_template("fund_details.html", form=form, fund=fund)
+
+
 @fund_bp.route("/create", methods=["GET", "POST"])
 def create_fund():
     """Creates a new fund"""
@@ -103,7 +113,7 @@ def create_fund():
     return render_template("fund.html", form=form, fund_id=None, error=error)
 
 
-@fund_bp.route("/<fund_id>", methods=["GET", "POST"])
+@fund_bp.route("/<uuid:fund_id>/edit", methods=["GET", "POST"])
 def edit_fund(fund_id):
     """Updates an existing fund"""
     fund = get_fund_by_id(fund_id)
@@ -146,7 +156,7 @@ def edit_fund(fund_id):
         update_fund(fund)
         if request.form.get("action") == "return_home":
             return redirect(url_for(INDEX_BP_DASHBOARD))
-        return redirect(url_for("fund_bp.view_fund", fund_id=fund.fund_id))
+        return redirect(url_for("fund_bp.view_fund_details", fund_id=fund.fund_id))
 
     error = error_formatter(form)
     return render_template("fund.html", form=form, fund_id=fund_id, error=error)
--- a/app/blueprints/fund/services.py
+++ b/app/blueprints/fund/services.py
@@ -1,3 +1,5 @@
+from flask import url_for
+
 from app.db.models import Fund
 
 
@@ -5,7 +7,11 @@ def build_fund_rows(funds: list[Fund]) -> list[dict]:
     rows = []
     for fund in funds:
         row = [
-            {"html": f"""<a class='govuk-link--no-visited-state'href='#'>{fund.name_json['en']}</a>"""},
+            {
+                # TODO move this html to template and use namespace functionality
+                "html": f"""<a class='govuk-link--no-visited-state'
+                href={url_for("fund_bp.view_fund_details", fund_id=fund.fund_id)}>{fund.name_json["en"]}</a>"""
+            },  # noqa: E501
             {"text": fund.description_json["en"]},
             {"text": fund.funding_type.get_text_for_display()},
         ]
--- a/app/blueprints/fund/templates/fund_details.html
+++ b/app/blueprints/fund/templates/fund_details.html
@@ -1,0 +1,161 @@
+{% extends "base.html" %}
+{% from "govuk_frontend_jinja/components/back-link/macro.html" import govukBackLink %}
+{%- from "govuk_frontend_jinja/components/summary-list/macro.html" import govukSummaryList -%}
+{% set showNavigationBar = True %}
+{% set active_item_identifier = "grants" %}
+{% block content %}
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+        {{ govukBackLink({
+          "text": "Back",
+          "href": url_for("fund_bp.view_all_funds")
+        }) }}
+        <h2 class="govuk-heading-l">{{ fund.name_json["en"] }}</h2>
+        {{ govukSummaryList({
+            "classes": "govuk-!-margin-bottom-9",
+                "rows": [
+                    {
+                        "key": {"text": "Grant name"},
+                        "value": {"text": fund.name_json["en"]},
+                        "actions": {
+                            "items": [
+                                {
+                                    "href": url_for(endpoint="fund_bp.edit_fund", fund_id=fund.fund_id, _anchor=form.name_en.id),
+                                    "text": "Change",
+                                    "classes": "govuk-link--no-visited-state",
+                                    "visuallyHiddenText": "Name english",
+                                }
+                            ]
+                        },
+                    },
+                    {
+                        "key": {"text": "Grant name (Welsh)"},
+                        "value": {"text": fund.name_json["cy"]},
+                        "actions": {
+                            "items": [
+                                {
+                                    "href": url_for(endpoint="fund_bp.edit_fund", fund_id=fund.fund_id, _anchor=form.name_cy.id),
+                                    "text": "Change",
+                                    "classes": "govuk-link--no-visited-state",
+                                    "visuallyHiddenText": "Name welsh",
+                                }
+                            ]
+                        },
+                    } if fund.welsh_available else {},
+                    {
+                        "key": {"text": "Short name"},
+                        "value": {"text": fund.short_name},
+                        "actions": {
+                            "items": [
+                                {
+                                    "href": url_for(endpoint="fund_bp.edit_fund", fund_id=fund.fund_id, _anchor=form.short_name.id),
+                                    "text": "Change",
+                                    "classes": "govuk-link--no-visited-state",
+                                    "visuallyHiddenText": "short name",
+                                }
+                            ]
+                        }
+                    },
+                    {
+                        "key": {"text": "Application heading"},
+                        "value": {"text": "Apply for " + fund.title_json["en"]},
+                        "actions": {
+                            "items": [
+                                {
+                                    "href": url_for(endpoint="fund_bp.edit_fund", fund_id=fund.fund_id, _anchor=form.title_en.id),
+                                    "text": "Change",
+                                    "classes": "govuk-link--no-visited-state",
+                                    "visuallyHiddenText": "title english",
+                                }
+                            ]
+                        }
+                    },
+                    {
+                        "key": {"text": "Application heading (Welsh)"},
+                        "value": {"text": "Apply for " + fund.title_json["cy"] if fund.title_json["cy"] else ""},
+                        "actions": {
+                            "items": [
+                                {
+                                    "href": url_for(endpoint="fund_bp.edit_fund", fund_id=fund.fund_id, _anchor=form.title_cy.id),
+                                    "text": "Change",
+                                    "classes": "govuk-link--no-visited-state",
+                                    "visuallyHiddenText": "title welsh",
+                                }
+                            ]
+                        },
+                    } if fund.welsh_available else {},
+                    {
+                        "key": {"text": "Funding type"},
+                        "value": {"text": fund.funding_type.get_text_for_display()},
+                        "actions": {
+                            "items": [
+                                {
+                                    "href": url_for(endpoint="fund_bp.edit_fund", fund_id=fund.fund_id, _anchor=form.funding_type.id),
+                                    "text": "Change",
+                                    "classes": "govuk-link--no-visited-state",
+                                    "visuallyHiddenText": "funding type",
+                                }
+                            ]
+                        }
+                    },
+                    {
+                        "key": {"text": "Welsh available"},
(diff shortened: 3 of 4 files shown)


app/blueprints/fund/routes.py
```diff
@@ -64,6 +64,62 @@ def view_fund():
...
+    ]
+
+    summary_data = [
+        {
+            "key": {"text": field["key"]},
+            "value": {"text": field["value"]},
+            "actions": {
+                "items": [
+                    {
+                        "href": f"""{url_for("fund_bp.edit_fund", fund_id=fund.fund_id)}#{field.get("id")}""",
```
nuwan-samarasinghe, line 111: Put a comment saying that # was added to go to the particular component.

app/blueprints/fund/routes.py
```diff
@@ -64,6 +64,62 @@ def view_fund():
...
+            "id": "description_en",
+            "key": "Description",
+            "value": fund.description_json["en"],
+            "visuallyHiddenText": "description",
+        },
+        {
+            "id": "ggis_scheme_reference_number",
+            "key": "GGIS scheme reference number",
+            "value": fund.ggis_scheme_reference_number,
+            "visuallyHiddenText": "GGIS scheme reference number",
```
nuwan-samarasinghe, line 100: shall we add `visually_hidden_text `

app/blueprints/fund/services.py
```diff
@@ -1,11 +1,16 @@
...
 
 
 def build_fund_rows(funds: list[Fund]) -> list[dict]:
     rows = []
     for fund in funds:
         row = [
-            {"html": f"""<a class='govuk-link--no-visited-state'href='#'>{fund.name_json['en']}</a>"""},
+            {
+                "html": f"""<a class='govuk-link--no-visited-state'href=
+            {url_for("fund_bp.view_grant_details",grant_id=fund.fund_id)}>{fund.name_json['en']}</a>"""
```
nuwan-samarasinghe, line 12: check the spaces may be run `pre-commit `  ?  seems like some spaces are needed

nuwan-samarasinghe (changes requested):

app/blueprints/fund/routes.py
gidsg: Can we use the `_anchor` `url_for` option instead? https://flask.palletsprojects.com/en/stable/api/#flask.Flask.url_for

NarenderRajuB: Updated

app/blueprints/fund/routes.py
```diff
@@ -104,7 +160,7 @@ def create_fund():
 
 
 @fund_bp.route("/<fund_id>", methods=["GET", "POST"])
-def edit_fund(fund_id):
+def edit_fund(fund_id, focus=None):
```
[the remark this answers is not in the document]
NarenderRajuB: Thanks for pointing this. Code updated

app/blueprints/fund/routes.py
NarenderRajuB: Added

app/blueprints/fund/routes.py
```diff
@@ -146,7 +207,7 @@ def edit_fund(fund_id):
         update_fund(fund)
         if request.form.get("action") == "return_home":
             return redirect(url_for(INDEX_BP_DASHBOARD))
-        return redirect(url_for("fund_bp.view_fund", fund_id=fund.fund_id))
+        return redirect(url_for("fund_bp.view_grant_details", grant_id=fund.fund_id))
```
nuwan-samarasinghe, line 210: Please update with the latest main in main 

```        match request.args.get("action") or request.form.get("action"):
            case "return_home":
                return redirect(url_for(INDEX_BP_DASHBOARD))
            case "grants_table":
                return redirect(url_for("fund_bp.view_all_funds"))
            case _:
                return redirect(url_for("round_bp.create_round", fund_id=new_fund.fund_id))

```

nuwan-samarasinghe (changes requested):

app/blueprints/fund/routes.py
NarenderRajuB: Hi @nuwan-samarasinghe  if makes send if we have multiple if... elseif to switch to 'match ..case' for more readable format .. but for single if else we can just use it  and here we have return after if too. if you still want me to do that happy to change here

nuwan-samarasinghe: No worries my bad

nuwan-samarasinghe (approved):
LGTM

app/blueprints/fund/routes.py
```diff
@@ -64,6 +64,67 @@ def view_fund():
...
+
+    summary_data = [
+        {
+            "key": {"text": field["key"]},
+            "value": {"text": field["value"]},
+            "actions": {
+                "items": [
+                    {
+                        # anchor added to set focus to field
+                        "href": f"""{url_for(
```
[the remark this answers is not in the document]
NarenderRajuB: @gidsg  updated the code. I missed it after we used _anchor

gidsg (dismissed):

app/blueprints/fund/routes.py
```diff
@@ -64,6 +64,63 @@ def view_fund():
     return render_template("fund_config.html", **params)
 
 
+@fund_bp.route("/details", methods=["GET"])
```
wjrm500, line 67: Not a _huge_ fan of the `/details` name - my preference would be for a routing pattern more like the below:

```
# List resources
GET /resources

# Show form to create new resource
GET /resources/new

# Create new resource
POST /resources

# Show single resource
GET /resources/{id}

# Show form to edit resource
GET /resources/{id}/edit

# Update resource
PUT /resources/{id}     # Full update
PATCH /resources/{id}   # Partial update

# Delete resource
DELETE /resources/{id}
```

Would've been good to discuss this with me before submitting the PR, but as this has now been approved by two others, this doesn't have to block merge 👍

app/blueprints/fund/routes.py
```diff
@@ -64,6 +64,63 @@ def view_fund():
...
+def view_grant_details():
```
wjrm500, line 68: The pattern we've been following in the code so far is using the term `grant` anywhere user-facing, but `fund` everywhere else, so this would be `view_fund_details` in that case. It's definitely a confusing halfway house we're in right now though, so no worries!

app/blueprints/fund/routes.py
```diff
@@ -64,6 +64,63 @@ def view_fund():
...
+    """
+    Renders grant details page
+    """
+    fund_id = request.args.get("grant_id")
+    fund = get_fund_by_id(fund_id)
+    fields = [
```
wjrm500, line 74: My first instinct is that it makes more sense to define all of this stuff inside the template itself, as it's very much presentation and not logic, plus it keeps this module much cleaner

Seems the other reviewers are happy with this though, so maybe I'm on my own 🥹

gidsg: Agree, ideally we shouldn't be hardcoding HTML in controllers if possible. Can we try refactoring this if possible so the HTML is generated in the template?

NarenderRajuB: Initially it was in template after i had review comments to move this to controller to reduce the template code and have more readable code moved to controller. Happy to move to template  @gidsg / @wjrm500 , if you are not happy with thi

gidsg: Given all the flip-flopping on this, in that case let's stick with this for now.

samuelhwilliams: I think if we've been flip-flopping and ended up with a preferred decision, we should implement that rather than concede?

This all feels like templating data - would be good to move it there 🙏  I appreciate it's a bit faffy but it feels like there's consensus now, and this is the most appropriate time to implement it.

NarenderRajuB: Updated code

app/blueprints/fund/routes.py
NarenderRajuB: Updated

app/blueprints/fund/routes.py
NarenderRajuB: refactored urls

app/blueprints/fund/services.py
```diff
@@ -1,11 +1,16 @@
...
     for fund in funds:
         row = [
-            {"html": f"""<a class='govuk-link--no-visited-state'href='#'>{fund.name_json['en']}</a>"""},
+            {
+                "html": f"""<a class='govuk-link--no-visited-state'
+                href={url_for("fund_bp.view_fund_details",fund_id=fund.fund_id)}>{fund.name_json['en']}</a>"""
+            },  # noqa: E501
             {"text": fund.description_json["en"]},
             {"text": fund.funding_type.get_text_for_display()},
         ]
```
gidsg, line 16: Could we also move this logic also into the template?

A trick to do this is to use the namespace functionality to build up the rows in the template e.g. as per example in: https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/2e304c2211a209ff52d09d28a5a323de6d3d99d1/submit/templates/submit/main/errorTable.html

wjrm500: Good idea - let me put up a separate PR for that as we should also apply this in the case of the templates table and remove `build_rows` in `app/blueprints/template/services.py`s

wjrm500: Actually after investigating further and discussing with @NarenderRajuB, this would require changes to the `GenericTablePage` class which we should perhaps de-prioritise for the time being, as the current implementation feels _good enough_. Definite TODO for the future though

wjrm500 (commented):
Looking good, couple of last changes then happy to approve...

- As per discussion around fund and grant terms yesterday, please rename `grant_details.html` to `fund_details.html`
- The fields displayed on the details page in [Figma](https://www.figma.com/design/H2YCtJp5KfYpqZLP34vWyg/FAB-(Fund-Application-Builder)?node-id=3919-31735&t=xSVgh6QxsjsagyIX-0) are slightly different to the fields displayed here - maybe it has changed in Figma since you implemented, or maybe you've had a conversation with Mika and the Figma needs to be updated - just pointing out the discrepancy

I think this implementation has exposed a minor design issue, with the fact that we don't mention anywhere on the page that we are dealing with a grant. I'll mention to Mika

wjrm500 (commented):
`active_item_identifier` should be "grants" not "Grants" otherwise it doesn't focus the navigation tab properly, see `navigation.html`. Always good to spin app up locally and double check any changes actually work, code can be a fickle thing 😸

nuwan-samarasinghe (approved):
LGTM

app/blueprints/fund/services.py
nuwan-samarasinghe: I think even flash messages we do. html may be need to consider about that also ?

app/blueprints/fund/templates/fund_details.html
```diff
@@ -0,0 +1,112 @@
...
+                "rows": [
+                    {
+                        "key": {"text": "Grant name"},
+                        "value": {"text": fund.name_json["en"]},
+                        "actions": {
+                            "items": [
+                                {
+                                    "href": url_for(endpoint="fund_bp.edit_fund", fund_id=fund.fund_id, _anchor="name_en"),
+                                    "text": "Change",
+                                    "visuallyHiddenText": "name_en",
```
samuelhwilliams, line 25: The visually hidden text here should not be the column/field name - it should be something that makes sense in a normal sentence.

So I think this should just be `grant name` here, `short name` below is fine, next should be `English title` rather than `title_en` - etc.

app/blueprints/fund/templates/fund_details.html
```diff
@@ -0,0 +1,112 @@
...
+        {{ govukSummaryList({
+            "classes": "govuk-!-margin-bottom-9",
+                "rows": [
+                    {
+                        "key": {"text": "Grant name"},
+                        "value": {"text": fund.name_json["en"]},
+                        "actions": {
+                            "items": [
+                                {
+                                    "href": url_for(endpoint="fund_bp.edit_fund", fund_id=fund.fund_id, _anchor="name_en"),
```
samuelhwilliams, line 23: The anchor tags here need to tie up to form fields on the `FundForm`. This is currently implicit and relies on the IDs being written correctly by a developer in each of these lines.

If we pass a `form` instance into this template, then we can grab the IDs of the fields programmatically which feels like a safer lookup.

Eg

```suggestion
                                    "href": url_for(endpoint="fund_bp.edit_fund", fund_id=fund.fund_id, _anchor=form.name_en.id),
```

app/blueprints/fund/templates/fund_details.html
```diff
@@ -0,0 +1,112 @@
...
+{% set showNavigationBar = True %}
+{% set active_item_identifier = "grants" %}
+{% block content %}
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+        {{ govukBackLink({
+          "text": "Back",
+          "href": url_for("fund_bp.view_all_funds")
+        }) }}
+        <h2 class="govuk-heading-m">{{ fund.name_json["en"] }}</h2>
```
samuelhwilliams, line 13: I think this should be a `govuk-heading-l`

samuelhwilliams (commented):
Just a couple questions to check with design/content, but the below dev comments for now please :)

I think we're close 🤞

app/blueprints/fund/templates/fund_details.html
NarenderRajuB: Updated

samuelhwilliams (approved):
Thanks for sticking with all the comments - I know there's been a lot of back and forth on this one 👍 We can do a final scan for content at the end and make sure everything's in sync around things like application heading vs grant title.

PR #207, opened 2025-01-14
Add and instrument flask-debugtoolbar

samuelhwilliams:
### Change description
This should help with local development, by making it easier to eg identify what SQL queries are run, see log messages, see rendered templates etc.

### How to test
* Re-build the `fab` docker compose service
* Run it
* Go through some of the FAB pages - open the toolbar on the right and you should be able to see SQL queries, etc.

fund-application-builder levellingup gov localhost_3011_grants__page=1

### DIFF
--- a/app/create_app.py
+++ b/app/create_app.py
@@ -61,6 +61,11 @@ def create_app() -> Flask:
         compare_server_default=True,
     )
 
+    if flask_app.config["FLASK_ENV"] == "development":
+        from flask_debugtoolbar import DebugToolbarExtension
+
+        DebugToolbarExtension(flask_app)
+
     # Configure application security with Talisman
     Talisman(flask_app, **Config.TALISMAN_SETTINGS)
 
--- a/config/envs/default.py
+++ b/config/envs/default.py
@@ -37,6 +37,7 @@ class DefaultConfig(object):
     SECURE_CSP = {
         "default-src": "'self'",
         "script-src": ["'self'"],  # Needed to enable nonce
+        "style-src": ["'self'"],
     }
 
     # Talisman Config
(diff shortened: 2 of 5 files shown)


app/create_app.py
```diff
@@ -61,6 +61,31 @@ def create_app() -> Flask:
         compare_server_default=True,
     )
 
+    if flask_app.config["FLASK_ENV"] == "development":
+        from flask_debugtoolbar import DebugToolbarExtension
+
+        DebugToolbarExtension(flask_app)
+
+        Config.TALISMAN_SETTINGS["content_security_policy"]["script-src"] += [
```
wjrm500, line 69: Having all of this config inside `create_app.py` doesn't seem right - can't we do here [what we do in pre-award repo](https://github.com/communitiesuk/funding-service-pre-award/blob/main/config/envs/development.py#L25-L40) and override these settings in development.py?

samuelhwilliams: yep 👍  Copied this from post-award, so I'll update that as well 👀

samuelhwilliams: https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/889

wjrm500:
Have tried this locally (after rebuilding) and I can't see any debug toolbar on the web page - on debugging I can see that the `FLASK_ENV` is definitely `development` and that `DevelopmentConfig` is definitely being loaded as the `Config`, but... yeah, something appears to be amiss, any ideas @samuelhwilliams?

samuelhwilliams:
@wjrm500 no idea what you mean 👀  (I've pushed a fix)

I assume you are totally OK working - you do not need to review any of this and I can find someone else :)

wjrm500 (approved):
Nice, working beautifully now. Very useful stuff

PR #215, opened 2025-01-16
Fix GOV.UK Service navigation implementation

samuelhwilliams:
Guidance for this component says that when it's used beneath a GOV.UK Header, we need to set the header border (the blue bar) to be full width.

https://design-system.service.gov.uk/components/service-navigation/

## Before
fund-application-builder levellingup gov localhost_3011_templates (1)

## After
fund-application-builder levellingup gov localhost_3011_templates

### DIFF
--- a/app/templates/base.html
+++ b/app/templates/base.html
@@ -13,6 +13,7 @@
 {% block header %}
   {{ govukHeader({
     "useTudorCrown": "yes",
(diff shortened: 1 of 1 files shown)


nuwan-samarasinghe:
Oh my god I really thought no change finally found the blue bar

nuwan-samarasinghe (approved):
LGTM

PR #219, opened 2025-01-17
flash loading with html template

nuwan-samarasinghe:
### Change description
adding flash message as a template

### How to test
Grant creation & Application creation will give the template

image

### DIFF
--- a/app/blueprints/fund/routes.py
+++ b/app/blueprints/fund/routes.py
@@ -94,7 +94,7 @@ def create_fund():
         add_fund(new_fund)
         flash_message(
             message="New grant added successfully",
-            href="#",
+            href=url_for("fund_bp.view_fund_details", fund_id=new_fund.fund_id),
             href_display_name=form.name_en.data,
         )
         match request.args.get("action") or request.form.get("action"):
--- a/app/blueprints/fund/routes.py
+++ b/app/blueprints/fund/routes.py
@@ -2,7 +2,6 @@ from datetime import datetime
 
 from flask import (
     Blueprint,
-    flash,
     redirect,
     render_template,
     request,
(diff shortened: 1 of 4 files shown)


app/blueprints/fund/routes.py
```diff
@@ -93,14 +92,11 @@ def create_fund():
...
-            <h3 class="govuk-notification-banner__heading">New grant added successfully</h3>
-            <p class="govuk-body">
-                <a class="govuk-notification-banner__link" href="#">
-                View {form.name_en.data}
-                </a>
-            </p>
-        """)
+        flash_message(
+            message="New grant added successfully",
+            href="#",
```
samuelhwilliams, line 97: Can we update this to the 'view grant' URL now? I know unrelated, but since we're touching the code and that page exists, might as well fix it here 🙏

samuelhwilliams (approved):
Nice 👍  One optional comment

app/shared/helpers.py
```diff
@@ -71,3 +72,35 @@ def all_funds_as_govuk_select_items(all_funds: list) -> list:
...
+        flash_message(
+            message="Your changes were saved successfully",
+            href="/dashboard",
+            href_display_name="Go to Dashboard",
+            next_href="/settings",
+            next_href_display_name="Edit Settings"
+        )
+        ```
+    """
+    flash(render_template("partials/flash_template.html", **locals()))
```
sfount, line 106: Purely for future code when we're going for something robust and out of "proof of concept". Heads up that the flash mechanism works by setting headers (sometimes cookies sometimes separate) that are then persisted through redirects back & forth to the users browser. 

A few risks I can think of for rendering a template that is then stored in a flash message: 
- The header size grows too big, causing the request to be rejected by firewalls/ browsers/ web apps
- Translation works inconsistently between the page being rendered and the flash message generated elsewhere
- It's hard to understand what could be shown on a page by looking at the templates - you would need to understand both the Python code and the templates to reason about the possible page states

When we come to bringing this code in my recommendation would be that we put as little information as possible in the flash message - maybe here that could be a code:
```python
flash("FUND_UPDATED_SUCCESSFULLY")
```

If it needed to, the template could then respond to those code and give the programmer a clear idea of what it could possibly show given different states:

```jinja
{% with messages = get_flashed_messages() %}

{% for message in messages %}

{% if message == "FUND_UPDATED_SUCCESSFULLY" %}
    {% set html %}
    <p class="govuk-notification-banner__heading">...</p>
    {% endset %}
{% else %}
    <!-- oops - default content -->
{% endif %}

{{ govukNotificationBanner({
    "html": html
}) }}
{% endfor %}

{% endwith %}
```

It could also be a small dictionary if additional context was needed by the template but the key thing would be that it would have a deterministic size that couldn't grow.

PR #223, opened 2025-01-20
Move some assets to top of static dir for GOV.UK Frontend

samuelhwilliams:
Caught in QA of https://mhclgdigital.atlassian.net/browse/FS-4947

### Change description
Precompiled GOV.UK Frontend html and css expects some assets (eg images, fonts, manifest) to be located at specific URL paths. We could only change this by bundling GOV.UK frontend ourselves, or doing some extremely gross search+replace in the bundled assets.

Instead of both of those, we can just lift the assets into the correct place for our local static files.

## Before
image

## After
image

### DIFF
--- a/build.py
+++ b/build.py
@@ -60,23 +60,12 @@ def build_govuk_assets(static_dist_root="app/static/dist"):
     for file_to_move in os.listdir(ASSETS_PATH):
         shutil.move("/".join([ASSETS_PATH, file_to_move]), DIST_PATH)
 
-    # Update relative paths
-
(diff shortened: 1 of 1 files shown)


sfount (commented):
Can this also remove this workaround for updating the baked in CSS paths manually? (haven't looked at the whole file or process!) https://github.com/communitiesuk/funding-service-design-fund-application-builder/blob/03564e2a1b96b91068165c75f353515383c0b1ec/build.py#L65-L79

sfount (approved):

PR #225, opened 2025-01-20
Bau/govuk frontend wtf

samuelhwilliams:
### Change description
Adds [govuk-frontend-wtf](https://github.com/LandRegistry/govuk-frontend-wtf) so that we can use this instead of writing our own macros for rendering form fields.

### DIFF
--- a/app/create_app.py
+++ b/app/create_app.py
@@ -5,6 +5,7 @@ from fsd_utils.authentication.decorators import SupportedApp, check_internal_use
 from fsd_utils.healthchecks.checkers import FlaskRunningChecker
 from fsd_utils.healthchecks.healthcheck import Healthcheck
 from fsd_utils.logging import logging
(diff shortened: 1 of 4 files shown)


app/create_app.py
```diff
@@ -75,11 +76,18 @@ def create_app() -> Flask:
     flask_app.jinja_loader = ChoiceLoader(
         [
             PackageLoader("app"),
-            PrefixLoader({"govuk_frontend_jinja": PackageLoader("govuk_frontend_jinja")}),
+            PrefixLoader(
+                {
```
nuwan-samarasinghe, line 80: What does this change do @samuelhwilliams ?

samuelhwilliams: It tells the flask app/jinja where to look for templates that are provided by the `govuk-frontend-wtf` library

nuwan-samarasinghe (approved):
nice work

PR #226, opened 2025-01-20
Consistently show flashed messages on all pages

samuelhwilliams:
Picked up in QA of https://mhclgdigital.atlassian.net/browse/FS-4925

### Change description
We use flash messages/notification banners to alert users of successful important actions.

We were handling these messages in each individual template, and only in specific templates, meaning that depending on which page you land on after a successful action, you may or may not get shown the flash message - it depends if that page is set up to show them.

This inconsistent behaviour is confusing and can lead to flash messages being shown much later, when they're no longer relevant.

Instead, let's show flash messages as part of the base template, meaning every page can show them by default.

### DIFF
--- a/app/blueprints/index/templates/dashboard.html
+++ b/app/blueprints/index/templates/dashboard.html
@@ -2,16 +2,8 @@
 {% set showNavigationBar = True %}
 {% set active_item_identifier = "home" %}
 
-{% from "govuk_frontend_jinja/components/notification-banner/macro.html" import govukNotificationBanner %}
 {% block content %}
     <div class="govuk-width-container">
-        {% with messages = get_flashed_messages() %}
-            {% if messages %}
-                <div class="govuk-grid-row">
-                    {{ govukNotificationBanner({"html": messages[0], "type": "success"}) }}
-                </div>
-            {% endif %}
-        {% endwith %}
         <div class="govuk-grid-row">
             <div class="govuk-grid-column-two-thirds">
                 <h1 class="govuk-heading-l">Creating a new grant application</h1>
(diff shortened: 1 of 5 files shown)


app/templates/base.html
```diff
@@ -34,6 +35,16 @@
 <div class="govuk-width-container {{ containerClasses }}">
   {% block beforeContent %}{% endblock beforeContent %}
   <main class="govuk-main-wrapper {{ mainClasses }}" id="main-content" role="main"{% if mainLang %} lang="{{ mainLang }}"{% endif %}>
+    {% with messages = get_flashed_messages() %}
+        {% if messages %}
+            <div class="govuk-grid-row">
+              <div class="govuk-grid-column-full">
```
sfount, line 41: I predict that if a page opts to use `govuk-grid-column-two-thirds` and this is full page width it will look funny but I suppose looking funny isn't an issue at the moment! 

A way to get around this might be to either `{% include %}` (or macro but comes with some closure/ scope implications) this as a one liner explicitly on pages where it would be expected (in the context of their column layout) or to have a standard two thirds column layout in the base that superclasses the `content` block meaning that content is consistent by default unless a page overrides it

Neither feel necessary right now just thinking through how this might look if we wanted to use this pattern longer term

app/blueprints/index/templates/dashboard.html
```diff
@@ -2,16 +2,7 @@
 {% set showNavigationBar = True %}
 {% set active_item_identifier = "home" %}
 
-{% from "govuk_frontend_jinja/components/notification-banner/macro.html" import govukNotificationBanner %}
 {% block content %}
-    <div class="govuk-width-container">
```
sfount, line 7: Nice catch

app/templates/base.html
samuelhwilliams: Sorry that this went in before you commented 🙏 

Good thoughts, plenty to think about if/when we polish this up more.

PR #227, opened 2025-01-20
Replace custom error handling with govuk-frontend-wtf

samuelhwilliams:
Let's remove some more of our custom code and rely on govuk-frontend-wtf to do the heavy and consistent lifting.

Visually everything should be unchanged, except some error messages which no longer have the <field_name> prefix. We have [a ticket to rationalise all error messages](https://mhclgdigital.atlassian.net/browse/FS-4967) later, so I'm ok with this.

### DIFF
--- a/app/blueprints/application/routes.py
+++ b/app/blueprints/application/routes.py
@@ -67,12 +67,9 @@ def select_fund():
     form.fund_id.choices = choices
     if form.validate_on_submit():
         return redirect(url_for("application_bp.select_application", fund_id=form.fund_id.data))
-    error = None
-    if form.fund_id.errors:
-        error = {"titleText": "There is a problem", "errorList": [{"text": form.fund_id.errors[0], "href": "#fund_id"}]}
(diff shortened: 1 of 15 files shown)


MarcUsher (commented):
Just about to go through the code a bit more but on checking out the branch and testing a bit the behaviour has changed quite a bit on eg. 'Add an application' with which fields it thinks is required:

Before:
Screenshot 2025-01-20 at 16 53 35


After:
image

Like you say in your PR comment the messages themselves will be updated as part of another ticket (which can add the updated copy for datetime field validation) but the new widget seems to be saying every field is required in the error summary where it's not in the actual form (this might be intentional and something to clean up in the other ticket but just wanted to highlight in case it wasn't!)

MarcUsher (approved):
Nice one, looks good!

@nuwan-samarasinghe FYI this will affect how error messages work on the Templates pages you're working on so worth having a look at how it's implemented here on the other pages!

PR #237, opened 2025-01-22
FS-4969 :Add and Edit grant template update with navigation

NarenderRajuB:
Ticket: https://mhclgdigital.atlassian.net/browse/FS-4969

Update Add Edit grant flows

### Change description

- [ x] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### How to test
Add a Grant
1. Changes Added to update Add a new grant and Update Grant templates use wtf forms
2. Click on Add a new grant from home/dashboard. Add details and click on continue and save. should redirect to create application/round page
3. Click on Add a new grant from select grant page. Add details and click on continue and save. should redirect to create application/round page
4. Click on Add a new grant from grants/funds list . Add details and click on continue and save. should redirect to create funds/grants list page.
5. Click on back/cancel should navigate to the page from where this operation is initiated
6. Click on save and return home will create a fund/grant and redirect to dashboard

Edit/Update Grant:

1. clicking on continue and save Edit/Update grant page will update the grant and redirect to the page where update is initiated.

image
imageimage

### DIFF
--- a/app/blueprints/fund/forms.py
+++ b/app/blueprints/fund/forms.py
@@ -1,7 +1,6 @@
-from enum import Enum
-
 from flask_wtf import FlaskForm
-from wtforms import HiddenField, RadioField, StringField, TextAreaField
+from govuk_frontend_wtf.wtforms_widgets import GovRadioInput, GovSubmitInput, GovTextArea, GovTextInput
+from wtforms import HiddenField, RadioField, StringField, SubmitField, TextAreaField
 from wtforms.validators import DataRequired, Length, ValidationError
 
 from app.db.models.fund import FundingType
@@ -16,37 +15,63 @@ def validate_unique_fund_short_name(form, field):
             raise ValidationError("Given fund short name already exists.")
 
 
-class GovUkRadioEnumField(RadioField):
-    source_enum: Enum
-    gov_uk_choices: list
-
-    def __init__(self, name: str, _prefix, _translations, label: str, _form, source_enum: Enum):
-        super().__init__(
-            name=name,
-            label=label,
-            _form=_form,
-            _prefix=_prefix,
-            _translations=_translations,
-            choices=[(value.name, value.value) for value in source_enum],
-        )
-        self.source_enum = source_enum
-        self.gov_uk_choices = [
-            {"text": value.get_text_for_display(), "value": value.value} for value in self.source_enum
-        ]
-
-
 class FundForm(FlaskForm):
     fund_id = HiddenField("Fund ID")
-    name_en = StringField("Grant name", validators=[DataRequired()])
-    name_cy = StringField("Grant name (Welsh)")
-    title_en = StringField("Application heading", validators=[DataRequired()])
-    title_cy = StringField("Application heading (Welsh)")
+    welsh_available = RadioField(
+        "Is this grant available in Welsh?",
+        choices=[("true", "Yes"), ("false", "No")],
+        coerce=lambda value: value == "true",
+        widget=GovRadioInput(),
+        default="false",
+        render_kw={"class": "govuk-radios govuk-radios--inline"},
+    )
+    name_en = StringField(
+        "Grant name",
+        widget=GovTextInput(),
+        description="For example, Community Ownership Fund",
+        validators=[DataRequired()],
+    )
+    name_cy = StringField(
+        "Grant name (Welsh)", widget=GovTextInput(), description="For example, Community Ownership Fund"
+    )
     short_name = StringField(
-        "Short name",
+        "Grant Short name",
+        widget=GovTextInput(),
+        description="A unique acronym of up to 10 characters for the grant.For example, COF",
         validators=[DataRequired(), Length(max=10), no_spaces_between_letters, validate_unique_fund_short_name],
     )
-    description_en = TextAreaField("Description", validators=[DataRequired()])
-    description_cy = TextAreaField("Description (Welsh)")
-    welsh_available = RadioField("Welsh available", choices=[("true", "Yes"), ("false", "No")], default="false")
-    funding_type = GovUkRadioEnumField(label="Funding type", source_enum=FundingType)
-    ggis_scheme_reference_number = StringField("GGIS scheme reference number", validators=[Length(max=255)])
+    title_en = StringField(
+        "Application heading",
+        widget=GovTextInput(),
+        description="For example, Apply for funding to save an asset in your community",
+        validators=[DataRequired()],
+    )
+    title_cy = StringField(
+        "Application heading (Welsh)",
+        widget=GovTextInput(),
+        description="For example, Apply for funding to save an asset in your community",
+    )
+    description_en = TextAreaField(
+        "Description",
+        widget=GovTextArea(),
+        description="What the grant is for. You can find this in the grant prospectus",
+        validators=[DataRequired()],
+    )
+    description_cy = TextAreaField(
+        "Description (Welsh)",
+        widget=GovTextArea(),
+        description="What the grant is for. You can find this in the grant prospectus",
+    )
+
+    funding_type = RadioField(
+        "Funding type",
+        choices=[(choice.value, choice.get_text_for_display()) for choice in FundingType],
+        widget=GovRadioInput(),
+    )
+    ggis_scheme_reference_number = StringField(
+        "GGIS scheme reference number",
+        widget=GovTextInput(),
+        validators=[Length(max=255)],
+    )
+    save_and_continue = SubmitField("Save and continue", widget=GovSubmitInput())
+    save_and_return_home = SubmitField("Save and return home", widget=GovSubmitInput())
--- a/app/blueprints/fund/routes.py
+++ b/app/blueprints/fund/routes.py
@@ -16,6 +16,10 @@ from app.shared.helpers import all_funds_as_govuk_select_items, flash_message
 from app.shared.table_pagination import GovUKTableAndPagination
 
 INDEX_BP_DASHBOARD = "index_bp.dashboard"
+SELECT_GRANT_PAGE = "select_grant"
+APPLICATIONS_DETAIL_PAGE = "view_application"
+FUND_LIST_PAGE = "grants_table"
+FUND_DETAILS_ROUTE = "fund_bp.view_fund_details"
 
 # Blueprint for routes used by v1 of FAB - using the DB
 fund_bp = Blueprint(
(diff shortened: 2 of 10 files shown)


app/blueprints/fund/templates/fund.html
```diff
@@ -17,39 +29,28 @@ <h1 class="govuk-heading-l">{{pageHeading}}</h1>
             <fieldset class="govuk-fieldset">
                 <form method="POST">
```
samuelhwilliams, line 30: Please disable HTML5 validation by adding `novalidate`

```suggestion
                <form method="POST" novalidate>
```

image

This is generally recommended practice: https://design-system.w3.org/styles/form-validation.html#:~:text=It%20is%20recommended%20to%20avoid,to%20the%20element.

app/blueprints/fund/forms.py
```diff
@@ -37,16 +38,62 @@ def __init__(self, name: str, _prefix, _translations, label: str, _form, source_
...
+        render_kw={"class": "govuk-radios govuk-radios--inline"},
+    )
+    name_en = StringField(
+        "Grant name",
+        widget=GovTextInput(),
+        description="For example, Community Ownership Fund",
+        validators=[DataRequired()],
+    )
+    name_cy = StringField(
+        "Grant name (Welsh)", widget=GovTextInput(), description="Er enghraifft, Gronfa Perchnogaeth Gymunedol"
```
samuelhwilliams, line 55: Let's keep instructions in English, and example content in Welsh please 🙏 

```suggestion
        "Grant name (Welsh)", widget=GovTextInput(), description="For example, Gronfa Perchnogaeth Gymunedol"
```

app/blueprints/fund/forms.py
```diff
@@ -37,16 +38,62 @@ def __init__(self, name: str, _prefix, _translations, label: str, _form, source_
...
+    description_en = TextAreaField(
+        "Description",
+        widget=GovTextArea(),
+        description="What the grant is for. You can find this in the grant prospectus",
+        validators=[DataRequired()],
+    )
+    description_cy = TextAreaField(
+        "Description (Welsh)",
+        widget=GovTextArea(),
+        description="Beth yw pwrpas y grant. Gallwch ddod o hyd i hwn ym mhrosbectws y grant",
```
samuelhwilliams, line 85: Service text should stay in English please

app/blueprints/fund/routes.py
```diff
@@ -16,6 +16,7 @@
 from app.shared.table_pagination import GovUKTableAndPagination
 
 INDEX_BP_DASHBOARD = "index_bp.dashboard"
+FUND_TEMPLATE = "fund.html"
```
samuelhwilliams, line 19: I know sonarqube flagged duplication, but just ignore that one and keep referencing the file inline - I think this is fine.

app/blueprints/fund/routes.py
```diff
@@ -91,21 +103,26 @@ def create_fund():
...
 
-    return render_template("fund.html", form=form, fund_id=None)
+        if form.save_and_return_home.data:
+            return redirect(url_for(INDEX_BP_DASHBOARD))
+
+        if request.args.get("action") == "grants_table":
+            return redirect(url_for("fund_bp.view_all_funds"))
+        return redirect(url_for("round_bp.create_round", fund_id=new_fund.fund_id))
+
+    return render_template(FUND_TEMPLATE, form=form, fund_id=None, prev_nav_url=prev_nav_url)
```
nuwan-samarasinghe, line 114: shall we pass all of these like **params 

`form=form, fund_id=None, prev_nav_url=prev_nav_url`

app/blueprints/fund/routes.py
```diff
@@ -91,21 +103,26 @@ def create_fund():
...
 
 
 @fund_bp.route("/<uuid:fund_id>/edit", methods=["GET", "POST"])
 def edit_fund(fund_id):
     """Updates an existing fund"""
     fund = get_fund_by_id(fund_id)
+    if request.args.get("actions") == "view_application":
```
nuwan-samarasinghe, line 121: shall we create a enum or a constant I think thats better for us I also even did not create but better to have these as a constant in one place `"view_application"`

app/blueprints/fund/routes.py
```diff
@@ -91,21 +103,26 @@ def create_fund():
...
+        round_id = request.args.get("round_id")
+        prev_nav_url = url_for("round_bp.round_details", round_id=round_id)
+    else:
+        prev_nav_url = url_for("fund_bp.view_fund_details", fund_id=fund_id)
```
nuwan-samarasinghe, line 125: Cant we think about to have these navigation identification in one place I cannot think a specific way if possible I think that will help us a lot

samuelhwilliams (commented):
This looks really good and seems consistent with designs/ticket etc - a few small comments - nice work 👍

app/blueprints/fund/templates/fund.html
```diff
@@ -17,39 +29,28 @@ <h1 class="govuk-heading-l">{{pageHeading}}</h1>
...
-                    {{input(form.title_en)}}
-                    {{input(form.title_cy, classes="welsh-field")}}
-                    {{input(form.short_name)}}
-                    {{multilineinput(form.description_en)}}
-                    {{multilineinput(form.description_cy, classes="welsh-field")}}
-                    {{radios_from_enum(form.funding_type)}}
-                    {{input(form.ggis_scheme_reference_number)}}
+                    {{ form.welsh_available(params={'classes':'govuk-radios--inline'} ) }}
+                    {{ form.name_en }}
+                    {{ form.name_cy(params= {"classes": "welsh-field"}) }}
```
nuwan-samarasinghe, line 34: Currently we use a java script to hide the welsh fields but its not ideal later we might need to change that

nuwan-samarasinghe (changes requested):

app/blueprints/fund/routes.py
samuelhwilliams: I think this comes down to personal preference; I actually prefer explicitly listing them rather than building a params dict and then doing **params - if we aren't reusing params anywhere, it seems like a layer of abstraction that doesn't give us anything useful

app/blueprints/fund/routes.py
samuelhwilliams: An enum, at least, feels like a good idea 👍

app/blueprints/fund/routes.py
samuelhwilliams: I think we could just split it out into a small helper function for each view for now.

ie instead of 

```python
@fund_bp.route("/<uuid:fund_id>/edit", methods=["GET", "POST"])
def edit_fund(fund_id):
    """Updates an existing fund"""
    fund = get_fund_by_id(fund_id)
    if request.args.get("actions") == "view_application":
        round_id = request.args.get("round_id")
        prev_nav_url = url_for("round_bp.round_details", round_id=round_id)
    else:
        prev_nav_url = url_for("fund_bp.view_fund_details", fund_id=fund_id)
```

we can have

```python
def _edit_fund_get_previous_url(action, fund_id, round_id):
    if request.args.get("actions") == "view_application":
        return url_for("round_bp.round_details", round_id=round_id)

    return url_for("fund_bp.view_fund_details", fund_id=fund_id)


@fund_bp.route("/<uuid:fund_id>/edit", methods=["GET", "POST"])
def edit_fund(fund_id):
    """Updates an existing fund"""
    fund = get_fund_by_id(fund_id)
    round_id = request.args.get("round_id")
    prev_nav_url = _edit_fund_get_previous_url(request.args.get("action"), fund_id, round_id)
```

This feels good and clear enough for now - don't need something more complex yet.

MarcUsher (commented):
Nice one Narender, looks good! Sam and Nuwan have made some good points, so just adding one small thing I spotted - now that you've updated the `FundForm` to use govukfrontend and wtforms widgets, we can lose the `GovUkRadioEnumField` class in that file which was only being used by the `funding_type` field in the form.

Using the `GovRadioInput` widget also means the error messages for the radio component now work properly and highlight the field in the form - nice one!

app/blueprints/fund/forms.py
```diff
@@ -37,16 +38,62 @@ def __init__(self, name: str, _prefix, _translations, label: str, _form, source_
 
 class FundForm(FlaskForm):
     fund_id = HiddenField("Fund ID")
-    name_en = StringField("Grant name", validators=[DataRequired()])
-    name_cy = StringField("Grant name (Welsh)")
-    title_en = StringField("Application heading", validators=[DataRequired()])
-    title_cy = StringField("Application heading (Welsh)")
+    welsh_available = RadioField(
+        "Is this grant available in Welsh?",
+        choices=[("true", "Yes"), ("false", "No")],
```
samuelhwilliams, line 43: When we do boolean radio fields, I think it's nice to add a `coerce` step that gives us true booleans:

```suggestion
        choices=[("true", "Yes"), ("false", "No")],
        coerce=lambda value: value == 'true',
```

Then in the view/route handler we can just check `form.welsh_available.data` for truthiness, rather than checking for `== "true"`

app/blueprints/fund/routes.py
```diff
@@ -72,11 +72,24 @@ def create_fund():
...
+        case "grants_table":
+            params["prev_nav_url"] = url_for("fund_bp.view_all_funds")
+        case "select_grant":
+            params["prev_nav_url"] = url_for("application_bp.select_fund")
+        case _:
+            params["prev_nav_url"] = url_for(INDEX_BP_DASHBOARD)
+
+    if request.method == "GET":
+        # Render the form on the initial GET request
+        return render_template("fund.html", **params)
```
samuelhwilliams, line 86: We can remove this altogether. `form.validate_on_submit()` already checks if it's a POST request. If it's not, then we will fall through to the end and do the render_template call there - which is identical. 🙏 

```suggestion
```

app/blueprints/fund/routes.py
NarenderRajuB: @samuelhwilliams i have reverted back now to use inline. we may need to pass this from Sonar

app/blueprints/fund/forms.py
```diff
@@ -16,37 +15,65 @@ def validate_unique_fund_short_name(form, field):
...
+        coerce=lambda value: value == "true",
+        widget=GovRadioInput(),
+        default="false",
+        render_kw={"class": "govuk-radios govuk-radios--inline"},
```
samuelhwilliams, line 26: I think this is redundant now?

```suggestion
```

app/templates/partials/flash_template.html
```diff
@@ -1,15 +1,15 @@
...
+    <a class="govuk-notification-banner__link--no-visited-state" href="{{ href }}">
         View {{ href_display_name }}
     </a>
 </p>
 {%- endif %}
 {%- if next_href and next_href_display_name %}
     <p class="govuk-body">
-        <a class="govuk-notification-banner__link" href="{{ next_href }}">
-            Next step: {{ next_href_display_name }}
+        Next step: <a class="govuk-notification-banner__link--no-visited-state" href="{{ next_href }}">
```
samuelhwilliams, line 11: I think we should be using `govuk-link` classes here for the right styling

```suggestion
        Next step: <a class="govuk-link govuk-link--no-visited-state" href="{{ next_href }}">
```

app/templates/partials/flash_template.html
```diff
@@ -1,15 +1,15 @@
 <h3 class="govuk-notification-banner__heading">{{ message }}</h3>
 {%- if href_display_name and href %}
 <p class="govuk-body">
-    <a class="govuk-notification-banner__link" href="{{ href }}">
+    <a class="govuk-notification-banner__link--no-visited-state" href="{{ href }}">
```
samuelhwilliams, line 4: And here

samuelhwilliams (approved):
A couple of tiny styling comments, otherwise functionality looks good to me - nice work 👍 

Approving as I trust you'll fix the styling issues. Once you have, go for it 👍

nuwan-samarasinghe (approved):
LGTM

PR #255, opened 2025-01-27
When next is empty, it raises a StopIteration since it cannot find an object which is happy with the condition so if it did not find any return None

nuwan-samarasinghe:
### Change description
When next is empty, it raises a StopIteration since it cannot find an object which is happy with the condition so if it did not find any return None

These errors happened due to a old bad data previously we had a feature to not to create a list component if there is a list with the same name so what happened was once you uploaded a wrong template Ex. half a list of local authorities and then if you upload the full list of local authorities then it will not update the list since name is already there so @wjrm500 did a change to fix and now even we have the list we create the list but unfortunately this test template was a older one before that fix so most of the values are not there so condition check fails and gives that error..

I suggest two work arounds 

1. delete data in test re create them or at lease get UAT dump  and put it into test
2. Upload fresh templates and use them for. testing

### DIFF
--- a/app/all_questions/read_forms.py
+++ b/app/all_questions/read_forms.py
@@ -25,9 +25,10 @@ def determine_display_value_for_condition(
         return "No" if lang == "en" else "Nac ydy"
     else:
         if list_name:
-            list_values = next(lizt["items"] for lizt in form_lists if lizt["name"] == list_name)
-            condition_text = next(item["text"] for item in list_values if item["value"] == condition_value)
-            return condition_text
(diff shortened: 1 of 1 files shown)


samuelhwilliams:
Can you explain how the error is happening? It's clear from the code and PR title what is going wrong, but not how it happens. That makes it hard to review and understand whether the fix is appropriate.

A reproduction test case would be v useful :pray:

nuwan-samarasinghe:
Updated the description of the PR

samuelhwilliams (approved):

PR #260, opened 2025-01-28
FS-4967: Adding error messages 

nuwan-samarasinghe:
**Ticket : https://mhclgdigital.atlassian.net/browse/FS-4967**

### Change description
Adding validation error messages 

### DIFF
--- a/app/blueprints/round/services.py
+++ b/app/blueprints/round/services.py
@@ -35,7 +35,7 @@ def build_round_rows(rounds: list[Round]) -> list[dict]:
                 "classes": "govuk-!-text-align-right fab-nowrap",
                 "html": f"""
                 <a class='govuk-link govuk-link--no-visited-state' href='{
-                    url_for("application_bp.build_application", round_id=round.round_id)
+                url_for("application_bp.build_application", round_id=round.round_id)
                 }'>Build application</a>""",
             },
         ]
@@ -46,6 +46,7 @@ def build_round_rows(rounds: list[Round]) -> list[dict]:
 def populate_form_with_round_data(round_obj, form_class):
     round_data = {
         "fund_id": round_obj.fund_id,
+        "welsh_available": round_obj.fund.welsh_available,
         "round_id": round_obj.round_id,
         "title_en": round_obj.title_json.get("en", ""),
         "title_cy": round_obj.title_json.get("cy", ""),
@@ -185,7 +186,8 @@ def update_existing_round(round_obj, form, user="dummy_user"):
     round_obj.eligibility_config = {"has_eligibility": form.eligibility_config.data}
     round_obj.eoi_decision_schema = {
         "en": convert_form_data_to_json(form.eoi_decision_schema_en.data),
-        "cy": convert_form_data_to_json(form.eoi_decision_schema_cy.data),
+        "cy": convert_form_data_to_json(
+            form.eoi_decision_schema_cy.data) if form.eoi_decision_schema_cy.data and form.eoi_decision_schema_cy.data.lower() != "none" else None,
     }
     round_obj.audit_info = {"user": user, "timestamp": datetime.now().isoformat(), "action": "update"}
     update_round(round_obj)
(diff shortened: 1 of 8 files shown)


app/shared/validators.py
```diff
@@ -17,3 +18,89 @@ def __call__(self, form, field):
...
+            # Optional path
+            r"(?:/[^/\s]*)*"
+            # Optional query string
+            r"(?:\?[^\s]*)?"
+            # Optional fragment
+            r"(?:#[^\s]*)?$"
+        )
+
+        if not re.match(pattern, field.data, re.IGNORECASE):
+            raise ValidationError(self.message)
```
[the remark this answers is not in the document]
nuwan-samarasinghe: when using that @Sam these will fail 

`
google.com-True
www.google.com-True
sub.google.com-True
google.co.uk-True
google.com/path-True
google.com/path/to/resource-True
google.com/path?query=value-True
google.com/path#fragment-True
google.com:8080-True
"-True
`
Dont we need to have this type of links ?

app/shared/validators.py
```diff
@@ -17,3 +18,89 @@ def __call__(self, form, field):
...
+
+    def __init__(self, message=None):
+        self.message = message
+
+    def __call__(self, form, field):
+        # Check if Welsh is available and name_cy is required
+        if isinstance(form.welsh_available.data, str):
+            form.welsh_available.data = True if form.welsh_available.data == "True" else False
+        if form.welsh_available and form.welsh_available.data and not (field.data and field.data.strip()):
+            raise ValidationError(self.message)
```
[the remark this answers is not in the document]
nuwan-samarasinghe: Added the change

app/shared/validators.py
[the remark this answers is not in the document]
samuelhwilliams: Ah, we've just moved this from another file - fair enough. Let's leave it for now 👍

app/shared/validators.py
```diff
@@ -17,3 +18,73 @@ def __call__(self, form, field):
...
+
+    def __init__(self, message=None):
+        self.message = message
+
+    def __call__(self, form, field):
+        str_content = field.data
+        if not str_content:
+            return
+        try:
+            json.loads(str_content)
```
samuelhwilliams, line 75: The only JSON we should accept is a dict/object, right? Not strings, numbers, lists?

nuwan-samarasinghe: https://github.com/communitiesuk/funding-service-design-fund-application-builder/blob/8a2a520b248e4ef9a9d396c0350158eb246c4b31/app/import_config/load_form_json.py#L335

so in this level we check wether are we getting a proper json but here we just check is it a correct json or not

samuelhwilliams: Something doesn't feel right there, it's letting me set it as a number

image

Let's do the validation at the form level. If `load_json_from_file` was doing deeper inspection it would still throw an exception, which would be a 500 rather than a nice form error?

nuwan-samarasinghe: let me check @samuelhwilliams

nuwan-samarasinghe: Fixed

app/blueprints/round/services.py
```diff
@@ -185,7 +186,8 @@ def update_existing_round(round_obj, form, user="dummy_user"):
     round_obj.eligibility_config = {"has_eligibility": form.eligibility_config.data}
     round_obj.eoi_decision_schema = {
         "en": convert_form_data_to_json(form.eoi_decision_schema_en.data),
-        "cy": convert_form_data_to_json(form.eoi_decision_schema_cy.data),
+        "cy": convert_form_data_to_json(
+            form.eoi_decision_schema_cy.data) if form.eoi_decision_schema_cy.data and form.eoi_decision_schema_cy.data.lower() != "none" else None,
```
samuelhwilliams, line 190: `form.eoi_decision_schema_cy.data.lower() != "none"`

does this really happen? 👀

PR #262, opened 2025-01-31
Bau/fix typo

samuelhwilliams:
### Change description
Fix rendering of the question text on 'show application' page.

While doing this I spotted that we were incorrectly rendering full labels instead of just the text for each field.

### Before
image
image

### After
image
image


MarcUsher (approved):
Thanks Sam, had fixed the rogue quote marks field name bit in my content PR that will be up soon but hadn't clocked the `.text` rendering issue, good fix! I'll rebase mine on top of this 👍

PR #266, opened 2025-01-31
FS-5004 : replicate browser back button using sessions

nuwan-samarasinghe:
**Ticket: https://mhclgdigital.atlassian.net/browse/FS-5004**

### Change description
Back button handling using sessions and here it will keep tack of the endpoints that we visit and accordingly it will go back once click on the back button in the UI

### DIFF
--- a/app/blueprints/index/routes.py
+++ b/app/blueprints/index/routes.py
@@ -9,6 +9,8 @@ from flask import (
     redirect,
     render_template,
     url_for,
+    session,
+    request
 )
 from fsd_utils.authentication.decorators import login_requested
 
@@ -42,6 +44,8 @@ def login():
 
 @index_bp.route("/dashboard", methods=["GET"])
 def dashboard():
+    session["visited_pages"] = []
+    session.modified = True
     return render_template("dashboard.html")
 
 
@@ -79,3 +83,17 @@ def download_form_json(form_id):
         mimetype="application/json",
         headers={"Content-Disposition": f"attachment;filename=form-{randint(0, 999)}.json"},  # nosec B311
     )
+
+
+@index_bp.route('/back')
+def go_back():
+    # If there are previous pages, pop the last one and redirect
+    if len(session['visited_pages']) > 1:
+        history = session['visited_pages']
+        history.pop()  # Remove the current page
+        prev_page = history[-1]
+        session["visited_pages"] = history
+        session.modified = True
+        return redirect(url_for(prev_page["endpoint"], **prev_page["view_args"], **prev_page["query_params"]))
+    else:
+        return redirect(url_for("index_bp.index"))  # If no previous page, go home
--- a/app/blueprints/index/routes.py
+++ b/app/blueprints/index/routes.py
@@ -44,8 +44,6 @@ def login():
 
 @index_bp.route("/dashboard", methods=["GET"])
 def dashboard():
-    session["visited_pages"] = []
-    session.modified = True
     return render_template("dashboard.html")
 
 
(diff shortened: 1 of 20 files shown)


gidsg:
Seems like quite a complex approach, as this is an internal tool have we considered using Javascript [Back API](https://developer.mozilla.org/en-US/docs/Web/API/History/back) instead?

nuwan-samarasinghe:
I also thought about it I will create a PR with that change also

app/shared/page_tracker.py
```diff
@@ -0,0 +1,69 @@
...
+                    return
+
+            # If the endpoint is not found, append it
+            visited_pages.append(page)
+            session.modified = True
+
+    def process_request(self, endpoint):
+        """Process the page tracking logic."""
+        self.initialize_session()
+        page = {"endpoint": endpoint, "view_args": request.view_args or {}, "query_params": request.args.to_dict()}
```
samuelhwilliams, line 67: Can I just check why you decided to use a dict here instead of storing `request.full_path`?

nuwan-samarasinghe: So using this endpoint it will simplify the usage of the check other wise we have to hardcode all the endpoints over here

so this dict will be having following data

"fund_bp.view_all_funds" and relevant params or resource id's

I hope we have to reduce the as much as possible of hardcoding @samuelhwilliams thats why I have used it

app/blueprints/index/routes.py
```diff
@@ -42,6 +44,8 @@ def login():
 
 @index_bp.route("/dashboard", methods=["GET"])
 def dashboard():
+    session["visited_pages"] = []
+    session.modified = True
```
samuelhwilliams, line 48: Do we need this or does stuff break? What happens if someone comes to FAB on a page that isn't the dashboard?

tests/shared/test_back_link.py
```diff
@@ -0,0 +1,80 @@
...
+        flask_test_client.get(url_for("template_bp.view_templates"))
+        assert len(session["visited_pages"]) == 1
+        assert session["visited_pages"][0]["endpoint"] == "template_bp.view_templates"
+
+        # Simulate visiting a page and reset once go into view_templates
+        _visit_template(flask_test_client, seed_dynamic_data)
+        assert len(session["visited_pages"]) == 3
+        flask_test_client.get(url_for("round_bp.view_all_rounds"))
+        assert len(session["visited_pages"]) == 1
+        assert session["visited_pages"][0]["endpoint"] == "round_bp.view_all_rounds"
```
samuelhwilliams, line 71: Please can we parameterise this rather than test three things in a single test.

I think parameters of `target_endpoint` should be fine?

```suggestion
@pytest.mark.parametrize("expected_endpoint", ["fund_bp.view_all_funds", "template_bp.view_templates", "round_bp.view_all_rounds"])
@pytest.mark.usefixtures("set_auth_cookie", "patch_validate_token_rs256_internal_user")
def test_session_track_visited_pages_reset_session(flask_test_client, seed_dynamic_data, expected_endpoint):
    with flask_test_client.session_transaction():
        # Simulate visiting a page and reset once go into view_all_funds
        _visit_template(flask_test_client, seed_dynamic_data)
        assert len(session["visited_pages"]) == 3
        flask_test_client.get(url_for(expected_endpoint))
        assert len(session["visited_pages"]) == 1
        assert session["visited_pages"][0]["endpoint"] == expected_endpoint
```

nuwan-samarasinghe: sure I will simplify this

app/blueprints/index/routes.py
nuwan-samarasinghe: Fixed

tests/shared/test_back_link.py
nuwan-samarasinghe: Fixed

samuelhwilliams (approved):

PR #272, opened 2025-02-04
FS-5006: improvements for the all questions pages

nuwan-samarasinghe:
**Ticket: https://mhclgdigital.atlassian.net/browse/FS-5005**

### Change description
In this PR we have mostly focused on following changes 

1. remove numbered references and give textual representation 
2. remove hard number system
3. simplify html generation for more readability

Templates's Questions View 
image

Application build question view 
image

All questions page view 
image


MarcUsher (commented):
Nice work Nuwan! Based on our discussion yesterday I know we said we're only doing minimal tweaks to these views for MVP as it's not crucial but just a few points below as design-wise we've still got a few discrepancies between each of the places that renders viewing questions (the first two I think are using the same method to render so changes to one should go to both). It might be nice to at least make these all consistent:

### Build application > View all application questions
image
- missing the navigation (see designs)
- the xl header for an h1 seems a bit overkill (but I guess we need it to differentiate with the L, M and S further down the page)?

### Build application > View questions (for single template)
image
- missing the navigation (see designs)
- the xl header for an h1 seems a bit overkill (but I guess we need it to differentiate with the L, M and S further down the page)?
- the field types seem to be showing on this view but no others
- this view is essentially the same as the one below (letting the user view questions of a single template) but the design of both pages is quite different eg. here we have a table of contents when it's just one template, which doesn't seem necessary? Also header styles, the [Preview], potentially missing the 'This template contains the following questions."

### Template details > view quesitons
image
- this one most closely matches Mika's intended design so is maybe the pattern to follow for the others (except for the 'view all application questions' still including the table of contents which everyone agreed was a helpful thing?)
- however the section breaks are full-width where the others are not (and also not on the designs)

I think the first `govuk-section-break` above each section could also maybe be dropped, as I think that's causing the double section break we see on the 'View all application questions' page?

nuwan-samarasinghe:
As per the discussion had with @MarcUsher we are going to do the changes as much as we can

MarcUsher (commented):
Nice work Nuwan! Just a couple of small comments:

- `View questions`/`View all applications questions` views (from the Build Application page) should both show the Navigation bar (they don't currently)
- `View questions` view (from Build Application) and `View template questions` view (from Template Details) are both looking super consistent, but the former is using the Task Name as the page title and the latter is using the Template Name. Would be good to just pick one and make that consistent across both? @mikaelmetthey do you have a preference which one this is?

Those are my only two notes, as we agreed we didn't want to make wholesale changes to the 'View all applications questions' view as this is also what's used in Apply, and I think these changes will make it consistent enough across all 3 views.

chaneldiep:
Good spot @MarcUsher, thanks! Seems sensible to go with template names at the top of both 'view questions' pages for consistency.

MarcUsher (approved):
Well done for sticking with this one Nuwan, approving ✅

PR #273, opened 2025-02-04
Enable slack deployment notifications

samuelhwilliams:
This differs from the `notify_slack` setting, which is more like `alert_slack_on_failed_deployments`.

I'll maybe rename this separately...

Related:
https://github.com/communitiesuk/funding-service-design-workflows/pull/265


nuwan-samarasinghe (approved):
LGTM

PR #284, opened 2025-02-11
backlink fix and use one backlink every place but passing a boolean if backlink is needed

nuwan-samarasinghe:
**Ticket: https://mhclgdigital.atlassian.net/browse/FS-4973**

### Change description
backlink fix and use one backlink every place but passing a boolean if backlink is needed

### DIFF
--- a/app/templates/base.html
+++ b/app/templates/base.html
@@ -34,11 +34,12 @@
 
 {% block main %}
 <div class="govuk-width-container {{ containerClasses }}">
-  {% block beforeContent %}{% endblock beforeContent %}
+  {% block beforeContent %}
+      {% if back_link %}
+         {{ govukBackLink({"text": "Back","href": url_for("index_bp.go_back")}) }}
+      {% endif %}
+  {% endblock beforeContent %}
   <main class="govuk-main-wrapper {{ mainClasses }}" id="main-content" role="main"{% if mainLang %} lang="{{ mainLang }}"{% endif %}>
-    {% if back_link %}
-     {{ govukBackLink({"text": "Back","href": url_for("index_bp.go_back")}) }}
-    {% endif %}
     {% with messages = get_flashed_messages() %}
         {% if messages %}
             <div class="govuk-grid-row">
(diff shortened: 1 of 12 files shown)


app/templates/base.html
```diff
@@ -35,6 +36,9 @@
 <div class="govuk-width-container {{ containerClasses }}">
   {% block beforeContent %}{% endblock beforeContent %}
   <main class="govuk-main-wrapper {{ mainClasses }}" id="main-content" role="main"{% if mainLang %} lang="{{ mainLang }}"{% endif %}>
+    {% if back_link %}
+     {{ govukBackLink({"text": "Back","href": url_for("index_bp.go_back")}) }}
+    {% endif %}
```
samuelhwilliams, line 41: The back link should go before the main class. See https://design-system.service.gov.uk/styles/layout/.

Please check the design system manual whenever you're making changes like this. It has a lot of useful guidance around how lay out the pages and where things go.

nuwan-samarasinghe: Yeah I also added but saw the figma and thought of adding it to the main but seems that wrong according to the https://design-system.service.gov.uk/styles/layout/ so added it to the correct place

app/templates/base.html
```diff
@@ -34,6 +35,9 @@
 {% block main %}
 <div class="govuk-width-container {{ containerClasses }}">
   {% block beforeContent %}{% endblock beforeContent %}
+  {% if back_link %}
+     {{ govukBackLink({"text": "Back","href": url_for("index_bp.go_back")}) }}
+  {% endif %}
```
samuelhwilliams, line 40: The backlink should be inside the beforeContent block, also

samuelhwilliams:
Please can we also go through and clean up the rest of the templates?

There are a lot that now have a redundant `beforeContent` block.

I also see some templates that still manually include a back link, eg build_application.html. Is this correct?  I feel like you've missed a few places.

app/templates/base.html
nuwan-samarasinghe: Thanks about this done the fix

samuelhwilliams:
Please can you tidy up all of the now-unused govukBacklink imports 🙏

nuwan-samarasinghe:
Fixed The changes

samuelhwilliams (approved):
nice ✅

PR #289, opened 2025-02-13
FS-4961: Backend pagination implementation and ellipsis addition when required 

nuwan-samarasinghe:
Ticket : https://mhclgdigital.atlassian.net/browse/FS-4961

### Change description
 Backend pagination implementation and ellipsis addition when required 

### How to test
Check the templates in UAT environment after deployment

image


app/blueprints/fund/routes.py
```diff
@@ -36,12 +35,18 @@ def view_all_funds():
...
+    pagination_json = pagination_convertor(pagination=pagination_data)
+    params = {
+        "table_pagination_page": {
+            "table": {
+                "table_header": [{"text": "Grant name"}, {"text": "Grant description"}, {"text": "Grant type"}],
+                "table_rows": build_fund_rows(pagination_data.items)
+            },
+            **({"pagination": pagination_json} if pagination_data else {})
+        }
+    }
```
samuelhwilliams, line 48: I think, between building this dict, and the dict in `pagination_convertor`, we're doing a lot of view-related stuff in Python.

Could we do this all in the templates?

It feels like we should be able to have a Jinja macro that takes a Flask-SQLAlchemy pagination object and displays the pagination component. And the headers and rows dicts could also be built up inside the templates.

I think the only thing we should need to pass into the template is the pagination object itself.

What do you think?

nuwan-samarasinghe: @samuelhwilliams it will not work because according to this logic we have to do some level of data conversion `pagination_convertor` we cannot send the paginator object it self and doing this in jinja is but costly I think

samuelhwilliams: (happy to pair and explore this together if helpful)

nuwan-samarasinghe: added the jinja templates as @samuelhwilliams mentioned

app/blueprints/round/services.py
```diff
@@ -19,7 +19,7 @@ def convert_form_data_to_json(data) -> dict:
     return {}
 
 
-def build_round_rows(rounds: list[Round]) -> list[dict]:
+def build_round_rows(rounds: list[dict]) -> list[dict]:
```
samuelhwilliams, line 22: I think we still want to avoid having this HTML in python files. Like with the pagination, can this go in each of the jinja files?

The `render_template` call should only need to be passed the pagination object, I think?

app/templates/macros/pagination.html
```diff
@@ -0,0 +1,45 @@
+{%- from "govuk_frontend_jinja/components/pagination/macro.html" import govukPagination -%}
+
+{% macro Pagination(params) %}
+    {% if params.data %}
+        {% set pagination = params.data %}
```
samuelhwilliams, line 5:
```suggestion
{% macro Pagination(pagination) %}
```

I don't think we need to call the arg `params` and have the require a `data` key - just call iit `pagination` directly and use that?

app/templates/macros/pagination.html
```diff
@@ -0,0 +1,45 @@
...
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-full">
+                {% if pagination.pages > 1 %}
+                    {% set prev_link = "?page=" ~ pagination.prev_num if pagination.has_prev else None %}
+                    {% set next_link = "?page=" ~ pagination.next_num if pagination.has_next else None %}
+
+                    {% set pagination_items = namespace(items=[]) %}
```
samuelhwilliams, line 12:
```suggestion
                    {% set pagination_items = [] %}
```

The namespace here feels overkill - can we just use a list and then replace all of `pagination_items.items` with `pagination_items`? Why do we need the extra layer?

app/templates/macros/pagination.html
nuwan-samarasinghe: let me try

nuwan-samarasinghe: Fixed

app/blueprints/round/services.py
nuwan-samarasinghe: Are we planing to do data transformations inside Jinja ? Definitly we have to move this transformation part from the python but my question is are we going to do this in Jinja better to check with the team because it seems we are moving some logic into jinja it violates the https://jinja.palletsprojects.com/en/stable/faq/?utm_source=chatgpt.com#isn-t-it-a-bad-idea-to-put-logic-in-templates

app/templates/macros/pagination.html
samuelhwilliams: I've seen the slightly wonky `{% set rows = rows.append() %}` syntax used frequently enough. I agree Jinja doesn't make it easy, but I think it's nicer to do the transformation from python objects to their HTML representation in one place - the jinja templates.

It feels fairly fragmented to have the python code generate some dicts that need to be in some very specific structure for the GOV.UK pagination component. That work happens really far away from where the pagination component is used, so it's harder to understand what is happening and why.

This is why I prefer to do that transformation in the jinja2 templates - because having the related things happen close together is easier to understand.

I agree that we should keep try to minimise the amount of _business_ logic inside jinja2 templates. But this is representation/display logic, and I think templates are very much a good place to have that.

app/blueprints/round/services.py
nuwan-samarasinghe: After the discussion with @samuelhwilliams planing to do a different PR for this change to minimise changes

samuelhwilliams (approved):
nice 👍

PR #294, opened 2025-02-21
FS-4961: Remove html python transformation for the table links into jinja form at and do data transformation inside jinja template for grants, application, template pages

nuwan-samarasinghe:
**Ticket: https://mhclgdigital.atlassian.net/browse/FS-4961**

### Change description
When we load the tables in each grant, application, template we do data transformation inside the python and it involves some html code so as a practice we should avoid using that so plan is to move those transformation into jinja format 

### How to test
Should work table pages as is


samuelhwilliams (approved):
Nice! This feels good.

PR #299, opened 2025-02-27
Adding ruff and pre-commit formatting changes 

nuwan-samarasinghe:
Adding ruff and pre-commit formatting changes since we have not enabled pre-commit there were lot of formatting issues but with this PR fixed all and @gidsg already added the check


wjrm500 (approved):
:star: :star: :star: :star: :star:

PR #308, opened 2025-03-03
FS-5033 - "Build application" page design changes

wjrm500:
### Ticket

[Button placement, UI clarity and navigation updates](https://mhclgdigital.atlassian.net/browse/FS-5033) (this PR only covers the third part of the ticket, around the "Build application" page)

### Description

- Replaced section action links with "Down" / "Up" links and "Edit" button, matching section.html styling
- Made task names clickable links opening previews in new tabs, removing separate "Preview" link

### Screenshot of UI change

image

### DIFF
--- a/app/blueprints/application/templates/build_application.html
+++ b/app/blueprints/application/templates/build_application.html
@@ -14,31 +14,42 @@
     </div>
     <ul class="app-task-list__items">
         {% for section in round.sections %}
-        <li class="">
+        <li class="task-list__new-design govuk-!-margin-bottom-2">
                 <span class="app-task-list__task-name">
                     <h3 class="govuk-heading-m">{{ section.index }}. {{ section.name_in_apply_json["en"] }}</h3>
                 </span>
                 <span class="app-task-list__task-actions">
-                    <a class="govuk-link--no-visited-state" href='{{ url_for("application_bp.section", round_id=round.round_id, section_id=section.section_id) }}'>Edit</a>
-                    <a class="govuk-link--no-visited-state govuk-!-margin-left-2" href='{{ url_for("application_bp.delete_section",round_id=round.round_id, section_id=section.section_id) }}'>Remove</a>
-                    {% if section.index > 1 %}
-                        <a class="govuk-link--no-visited-state govuk-!-margin-left-2" href='{{ url_for("application_bp.move_section_up_route",round_id=round.round_id, section_id=section.section_id) }}'>Move up</a>
+                    {% if section.index == round.sections | length %}
+                        <span class="govuk-!-font-size-19 govuk-!-margin-right-2 disabled-link">Down</span>
+                    {% else %}
+                        <a class="govuk-!-font-size-19 govuk-!-margin-right-2 govuk-link--no-visited-state"
+                           href='{{ url_for("application_bp.move_section_down_route",round_id=round.round_id, section_id=section.section_id) }}'>Down</a>
                     {% endif %}
-                    {% if section.index < round.sections | length %}
-                        <a class="govuk-link--no-visited-state govuk-!-margin-left-2" href='{{ url_for("application_bp.move_section_down_route",round_id=round.round_id, section_id=section.section_id) }}'> Move down</a>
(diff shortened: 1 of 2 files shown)


app/blueprints/application/templates/build_application.html
```diff
@@ -14,31 +14,42 @@ <h1 class="govuk-heading-l">Build application</h1>
...
+                                       target="_blank"
+                                       href='{{ url_for("index_bp.preview_form", form_id=form.form_id) }}'>
+                                        {{ form.name_in_apply_json["en"] }} (previews in a new tab)
+                                    </a>
                                 </h3>
                             </span>
                             <span class="app-task-list__task-actions">
-                                <a class="govuk-link--no-visited-state" href="{{ url_for('application_bp.view_form_questions', round_id=round.round_id, section_id=section.section_id, form_id=form.form_id) }}">View questions</a>
-                                <a class="govuk-link--no-visited-state govuk-!-margin-left-2" target="_blank" href='{{ url_for("index_bp.preview_form", form_id=form.form_id) }}'>Preview</a>
+                                <a class="govuk-link--no-visited-state govuk-!-font-size-19" href="{{ url_for('application_bp.view_form_questions', round_id=round.round_id, section_id=section.section_id, form_id=form.form_id) }}">View questions</a>
```
sfount, line 52: Not at all a big deal (especially as this follows existing precedent) but the appropriate way to style a link is to provide both `govuk-link` and the no visit state modifier.

```html
<a class="govuk-link govuk-link--no-visited-state"></a>
```

The `govuk-link--no-visited-state` is responsible for:
- styling the colour of the link

Everything else comes from the `govuk-link` class:
- font size
- responsive padding/ margins/ sizing
- interaction with other classes
- etc.

I _believe_ that the default page font size is the same as the `govuk-!-font-size-19` override so I think if you replaced where you've found you have to do that with `govuk-link` as well you should get an appropriately formatted link that should behave consistently with the rest of the design system classes.

wjrm500: Thanks Steven - when I replace `govuk-!-font-size-19` with `govuk-link` I get a computed font size of 16px, I believe this is the default

sfount: 🤔 curious! 19 is the default for `govuk-body` (not small `govuk-body-s`) and 16 is definitely the smallest it can go which I don't think is default https://design-system.service.gov.uk/styles/type-scale/#responsive-behaviour

I wonder if there's something fighting against us that we've set

sfount: aha https://github.com/communitiesuk/funding-service-design-fund-application-builder/blob/144ac34c8b00f75181683dbd50e7e39bcbf1eff6/app/static/src/styles/fab.css#L24

This font size override won't be responsive and there's a few others bits to consider with that.

I wonder if we have to keep pushing things in here back up to the 19 default if we should remove that and only apply `govuk-body-s` (`govuk-!-font-size-16`) on the elements that need it.

wjrm500: Ah yes sorry should have noted in my original reply - I did also remove the 16px set by the `.app-task-list__items` class in my original test, and we still get defaulted to a font-size of 16px because after that we still fall back to the 16px browser standard - see screenshot of me manipulating the CSS in the Dev Tools:

image

nuwan-samarasinghe (approved):
LGTM

app/blueprints/application/templates/build_application.html
sfount: After a bit of digging with me and @wjrm500 the going theory is that the browser default font size is also 16px where the design system default is slightly larger at 16px.

Everything here should ideally be wrapped in a `govuk-body` context near the top so that it has consistent size instructions going down, we can then only apply overrides where we really need the font to be smaller.

This goes along with ideally not adding our own fixed font overrides in the styling as that can lead to inconsistent behaviours we need to keep overriding like here! 

A sensible proposal to out-of-scope this from this PR and do a bit of a CSS/ structure audit separately and consistently 🙏

PR #310, opened 2025-03-04
FS-5038: TC-003 -> Add new application and back to dashboard

nuwan-samarasinghe:
**Ticket: https://mhclgdigital.atlassian.net/browse/FS-5038**

### Change description
FS-5038: TC-003 -> Add new application and back to dashboard

### How to test
Use steps in https://github.com/communitiesuk/funding-service-design-fund-application-builder/blob/main/tests/e2e/README.MD


tests/e2e/features/test_e2e_add_new_application_and_back_to_dashboard.py
```diff
@@ -0,0 +1,33 @@
...
+    (
+        DashboardPage(page, domains.fab_url)
+        .when_goto_dashboard()
+        .then_click_set_up_a_new_application()
+        .then_select_a_grant(grant_name=created_grant["grant_name"])
+        .then_click_continue()
+        .then_fill_application_details()
+        .then_click_save_and_return_home()
+        .and_validate_application_success_message()
+    )
```
samuelhwilliams, line 19: I am not sure about this approach - using fixtures makes this feel more like a unit test, but we are writing end-to-end tests. This is a fragment of a user journey, not a whole user journey. If the create-a-grant journey is broken, this test will also be broken. It does not work in isolation and I don't think we should pretend that it does.

I would be more inclined to a single, long end-to-end test that does all of: create a grant, create an application, build it, export it.

nuwan-samarasinghe: yeah I also thought about it but again thought team will ask for a fixture thats why I added that so I will go for my previous idea then @samuelhwilliams

samuelhwilliams: I'm also happy to defer to @wjrm500 and let him make the decisions on this work as I should have stepped away 😅

wjrm500: Hello both, I feel that this particular test is actually okay as it is, as it captures one of the expected user journeys end-to-end - that of a user coming to FAB after a grant has already been created, and continuing the flow at a later step via the dashboard. I think the current approach with the fixture is best as it means the test directly models the journey of this imaginary user, rather than mixing in a precursor journey (that of actually creating the grant) that could have been carried out by another user.

We will also have a separate test where we model the "Save and continue" flow, whereby the user starts on the dashboard, creates a grant, clicks "Save and continue", creates an application, clicks "Save and continue", builds the application from templates, and exports. I agree it's critical that we have this!

wjrm500: Although, in saying that, I've just realised that this test terminates after the application is created, without continuing to build the application from templates, which is less E2E-feeling isn't it... I think maybe there are some issues with the way these journeys have been defined, that extend beyond the scope of this PR (which does actually fulfil the brief).

Ultimately as long as we end up with an E2E test that does model the entire flow, as mentioned in the above comment, then I'm happy 👍

wjrm500 (changes requested):
First time I've looked properly at your E2E work @nuwan-samarasinghe and it's a joy to behold, I have to admit! When this command `uv run pytest --e2e --e2e-env dev --e2e-aws-vault-profile dev` just worked out of the box I was very impressed, and the test code is incredibly readable thanks to the `Page` abstraction - excellent stuff 🐈 

I do have some comments as discussed in-person just now:
- Current use of `when` / `then` prefixes in test flows is a bit unintuitive - almost all are actions which should be prefixed `when`, with `then` used to validate outcomes
- With this approach, we realise that we probably need more outcome validation (`then`) steps, so if we could add some of these, that would be great
- Could we add an extra test case for the user clicking "Save and continue" and returning to "Applications table" page

tests/e2e/conftest.py
```diff
@@ -134,6 +135,23 @@ def user_auth(
...
+    """
+    create_grant = (
+        DashboardPage(page, domains.fab_url)
+        .given_user_is_on_dashboard()
+        .when_click_add_a_new_grant()
+        .when_fill_non_welsh_competitive_grant_details()
+        .when_click_save_and_return_home(return_self=True)
+    )
+
+    return {"grant_name": create_grant.grant_name}
```
wjrm500, line 152: Best to return the grant itself here as seems more of an intuitive return object, then in the tests we can use `created_grant.grant_name` instead of `created_grant["grant_name"]`

tests/e2e/features/test_e2e_add_new_application_and_back_to_dashboard.py
```diff
@@ -0,0 +1,63 @@
...
+def test_add_new_application_from_dashboard_view(page: Page, domains: FabDomains, user_auth, created_grant):
+    (
+        DashboardPage(page, domains.fab_url)
+        .given_user_is_on_dashboard()
+        .then_verify_on_dashboard()
+        .when_click_set_up_a_new_application()
+        .then_verify_on_select_grant()
+        .when_select_a_grant(grant_name=created_grant["grant_name"])
+        .when_click_continue()
+        .then_verify_on_create_application()
```
wjrm500, line 18: What about...

```
...
 .then_verify_on_create_application()
 .then_verify_grant_name_visible_on_application(grant_name=created_grant["grant_name"])
...
```

Just a suggestion, if you feel like you want to implement please also do this for the other tests in this module.

tests/e2e/features/test_e2e_add_new_application_and_back_to_dashboard.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+
+# TC-003-[2]: Add application from applications page and comeback dashboard
+def test_add_application_from_applications_page_and_goback_dashboard(
+    page: Page, domains: FabDomains, user_auth, created_grant
+):
+    (
+        ApplicationsPage(page, domains.fab_url)
+        .when_goto_applications()
```
wjrm500, line 32: Should this be `given_user_is_on_applications` to follow naming convention? Please also correct for other test

tests/e2e/features/test_e2e_add_new_application_and_back_to_dashboard.py
```diff
@@ -0,0 +1,63 @@
...
+        .then_verify_on_application()
```
wjrm500, line 33: Should be `.then_verify_on_applications()` (plural) (please also correct for other test)

tests/e2e/features/test_e2e_add_new_application_and_back_to_dashboard.py
```diff
@@ -0,0 +1,63 @@
...
+        .when_click_create_new_application()
+        .then_verify_on_select_grant()
+        .when_select_a_grant(grant_name=created_grant["grant_name"])
+        .when_click_continue()
+        .then_verify_on_create_application()
+        .when_fill_application_details()
+        .when_click_save_and_continue()
+        .then_verify_on_application()
+        .and_validate_application_success_message()
```
wjrm500, line 62: You mentioned that there were different application success messages depending on if you returned to the dashboard versus back to the applications table - if this is true, then shouldn't we use different message validation methods across the two tests? (Currently both use `and_validate_application_success_message`)

tests/e2e/pages/applications_page.py
```diff
@@ -0,0 +1,32 @@
...
+        return self
+
+    def when_click_create_new_application(self):
+        self.create_new_application.click()
+        from tests.e2e.pages.select_grant_page import SelectGrantPage
+
+        return SelectGrantPage(self.page)
+
+    def then_verify_on_application(self):
+        expect(self.title).to_be_visible()
```
wjrm500, line 25: ❤️ this syntax

tests/e2e/pages/applications_page.py
```diff
@@ -0,0 +1,32 @@
...
+    def when_goto_applications(self):
+        if self.base_url:
+            self.page.goto(f"{self.base_url}/rounds")
+        return self
+
+    def when_click_create_new_application(self):
+        self.create_new_application.click()
+        from tests.e2e.pages.select_grant_page import SelectGrantPage
+
+        return SelectGrantPage(self.page)
```
wjrm500, line 22: Is it slightly confusing that we are passing the `ApplicationsPage` object to the constructor of `SelectGrantPage`, to be set as the `page` attribute (via `PageBase` constructor)? As this means if I was inspecting the code I'd see that my `SelectGrantPage.page` is an `ApplicationsPage` object - maybe it should be called `previous_page`? Not sure, what are your thoughts?

tests/e2e/pages/applications_page.py
```diff
@@ -0,0 +1,32 @@
...
+        self.create_new_application = self.page.get_by_role("button", name="Create new application")
+
+    def when_goto_applications(self):
+        if self.base_url:
+            self.page.goto(f"{self.base_url}/rounds")
+        return self
+
+    def when_click_create_new_application(self):
+        self.create_new_application.click()
+        from tests.e2e.pages.select_grant_page import SelectGrantPage
```
wjrm500, line 20: Any particular reason this import and others aren't at the module level? If not, I'd instinctively prefer to see all of the imports across all of the pages moved to the top of the page

tests/e2e/pages/create_application_page.py
```diff
@@ -0,0 +1,73 @@
...
+        self.save_and_return_home.click()
+        from tests.e2e.pages.dashboard_page import DashboardPage
+
+        return DashboardPage(self.page)
+
+    def when_click_save_and_continue(self):
+        self.save_and_continue.click()
+        from tests.e2e.pages.applications_page import ApplicationsPage
+
+        return ApplicationsPage(self.page)
```
wjrm500, line 55: This doesn't feel quite right - although it works in the context of the existing test, remember that if we click "Save and continue" when coming from the dashboard, we'd expect to be taken to the "Build application" page, not the "Applications" page. We'll have to fix this for our next test case PR anyway so might as well fix in this one

tests/e2e/pages/select_grant_page.py
```diff
@@ -0,0 +1,45 @@
...
+            selected_grant = random.choice(grant_options)
+        else:
+            matching_grants = [
+                option
+                for option in grant_options
+                if grant_name.lower() in option["label"].lower() or grant_name.lower() in option["value"].lower()
+            ]
+            if not matching_grants:
+                print(f"Warning: Grant '{grant_name}' not found. Selecting a random grant.")
+                selected_grant = random.choice(grant_options)
```
wjrm500, line 33: Do we want to just raise a warning here, or do we want the test to fail? Feel like we want the test to fail if a matching grant is not found?

tests/e2e/features/test_e2e_add_new_grant_and_back_to_dashboard.py
```diff
@@ -9,21 +9,27 @@
 def test_add_new_grant_from_dashboard_and_back_to_dashboard(page: Page, domains: FabDomains, user_auth):
     (
         DashboardPage(page, domains.fab_url)
-        .when_goto_dashboard()
-        .then_click_add_a_new_grant()
-        .then_fill_non_welsh_competitive_grant_details()
-        .then_click_save_and_return_home()
-        .and_validate_success_message()
+        .given_user_is_on_dashboard()
```
wjrm500, line 12: Understand these changes, but would ideally like to see any changes related to test case TC-002 at least in a separate commit in this PR, if not in an entirely separate PR. This would have made the changeset easier to follow. No worries at this stage, just a reminder to try and keep things separate to help the reviewer

wjrm500 (commented):
I've left a few comments, nothing major. I won't block merge just yet, but I'll also withhold approval until you've addressed a couple of my questions. Great work BTW!

tests/e2e/features/test_e2e_add_new_application_and_back_to_dashboard.py
wjrm500: Ignore me

tests/e2e/pages/applications_page.py
wjrm500: Ignore me again - the `page` attribute will always be the base page.

tests/e2e/pages/applications_page.py
wjrm500: Because of circular dependencies.

tests/e2e/pages/create_application_page.py
wjrm500: Have agreed will do this in next test case PR

tests/e2e/features/test_e2e_add_new_grant_and_back_to_dashboard.py
nuwan-samarasinghe: Sure as discuss later commits I will

tests/e2e/conftest.py
nuwan-samarasinghe: Fixed

tests/e2e/features/test_e2e_add_new_application_and_back_to_dashboard.py
nuwan-samarasinghe: added

tests/e2e/features/test_e2e_add_new_application_and_back_to_dashboard.py
nuwan-samarasinghe: fixed

tests/e2e/features/test_e2e_add_new_application_and_back_to_dashboard.py
nuwan-samarasinghe: Fixed

wjrm500 (approved):
Lovely!

PR #315, opened 2025-03-05
FSPT-228: Move test to communities domain name

gidsg:
### Change description
Config changes to move test to communities domain name

### How to test
After deployment service should be available on the test.communities.gov.uk domain name

### DIFF
--- a/copilot/fsd-fund-application-builder/manifest.yml
+++ b/copilot/fsd-fund-application-builder/manifest.yml
@@ -112,7 +112,7 @@ environments:
         port: 8080
   uat:
     variables:
-      FORM_DESIGNER_EXTERNAL_HOST: "https://form-designer.test.access-funding.test.levellingup.gov.uk"
+      FORM_DESIGNER_EXTERNAL_HOST: "https://form-designer.access-funding.test.communities.gov.uk"
     count:
       range: 2-4
       cooldown:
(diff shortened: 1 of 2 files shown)


nuwan-samarasinghe (approved):
LGTM

wjrm500 (commented):
Looks good, have a couple Qs:

- Why are we adding this `alias` thing - is it just for backwards compatibility?
- FAB's UAT environment points to Form Designer's Test environment (because Form Designer has never been deployed to UAT), so if we are changing Form Designer's domain name in Test as part of this suite of changes, we should update this too (line 115 in manifest.yml)

gidsg:
@wjrm500 
>  Why are we adding this alias thing - is it just for backwards compatibility?

Yes, this is a slightly janky workaround. Since all the DNS for the new and old domain names is pointing to the same CloudFront distribution, we need to keep these aliases, so that we can redirect in future from the old to new URLs i.e. when this function is enabled.

If we remove them they'll be removed from the CloudFront distribution aliases and therefore not resolve at all.

gidsg:
@wjrm500 

> FAB's UAT environment points to Form Designer's Test environment (because Form Designer has never been deployed to UAT), so if we are changing Form Designer's domain name in Test as part of this suite of changes, we should update this too (line 115 in manifest.yml)

Good spot, have updated in 1438641

Tiny49 (approved):
LGTM

wjrm500 (approved):

PR #316, opened 2025-03-05
Change URL fields in "Create an application" to have type=url

wjrm500:
### Description

Using `url` as the HTML element type for URL fields on the "Create an application" page allows tools like [Fake Filler](https://chromewebstore.google.com/detail/fake-filler/bnjjngeaknajbdcgpfkgnonkmififhfo) to correctly insert valid URLs instead of generic text, saving time during testing.


nuwan-samarasinghe (approved):
nice change :)

PR #322, opened 2025-03-07
When a user created the template and clicked on save and return home should go back to dashboard instead, it comes back to the template page

nuwan-samarasinghe:
### Change description
When a user created the template and clicked on save and return home should go back to dashboard instead, it comes back to the template page


NarenderRajuB (approved):
LGTM

PR #329, opened 2025-03-11
FS-5036: Fix Clone Lizt object for cloned components

NarenderRajuB:
Ticket : https://mhclgdigital.atlassian.net/browse/FS-5036

### Change description

- [ x] README and other documentation has been updated / added (if needed)
- [ ]x Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### DIFF
--- a/app/db/queries/clone.py
+++ b/app/db/queries/clone.py
@@ -1,7 +1,14 @@
 from uuid import uuid4
 
 from app.db import db
-from app.db.models import Component, Form, Page, Round, Section
+from app.db.models import Component, Form, Lizt, Page, Round, Section
+
+
+def _initiate_cloned_lizt(to_clone: Lizt) -> Lizt:
+    clone = Lizt(**to_clone.as_dict())
+    clone.list_id = uuid4()
+    clone.is_template = False
+    return clone
 
 
 def _initiate_cloned_component(to_clone: Component, new_page_id=None, new_theme_id=None):
@@ -57,6 +64,7 @@ def clone_single_section(section_id: str, new_round_id=None) -> Section:
     cloned_forms = []
     cloned_pages = []
     cloned_components = []
+    cloned_lizts = []
     # loop through forms in this section and clone each one
     for form_to_clone in section_to_clone.forms:
         cloned_form = _initiate_cloned_form(form_to_clone, clone.section_id, section_index=form_to_clone.section_index)
@@ -64,14 +72,16 @@ def clone_single_section(section_id: str, new_round_id=None) -> Section:
         for page_to_clone in form_to_clone.pages:
             cloned_page = _initiate_cloned_page(page_to_clone, new_form_id=cloned_form.form_id)
             cloned_pages.append(cloned_page)
-            # clone the components on this page
-            cloned_components.extend(
-                _initiate_cloned_components_for_page(page_to_clone.components, cloned_page.page_id)
+            # clone the components and lizts for component on this page
+            cloned_component_list_for_page, cloned_lizts_for_page = _initiate_cloned_components_for_page(
+                page_to_clone.components, cloned_page.page_id
             )
+            cloned_components.extend(cloned_component_list_for_page)
+            cloned_lizts.extend(cloned_lizts_for_page)
 
         cloned_forms.append(cloned_form)
 
-    db.session.add_all([clone, *cloned_forms, *cloned_pages, *cloned_components])
+    db.session.add_all([clone, *cloned_forms, *cloned_pages, *cloned_components, *cloned_lizts])
     cloned_pages = _fix_cloned_default_pages(cloned_pages)
     db.session.commit()
 
@@ -100,11 +110,16 @@ def clone_single_form(form_id: str, new_section_id=None, section_index=0) -> For
 
     cloned_pages = []
     cloned_components = []
+    cloned_lizts = []
     for page_to_clone in form_to_clone.pages:
         cloned_page = _initiate_cloned_page(page_to_clone, new_form_id=clone.form_id)
         cloned_pages.append(cloned_page)
-        cloned_components.extend(_initiate_cloned_components_for_page(page_to_clone.components, cloned_page.page_id))
-    db.session.add_all([clone, *cloned_pages, *cloned_components])
+        cloned_component_list_for_page, cloned_lizt_per_page = _initiate_cloned_components_for_page(
+            page_to_clone.components, cloned_page.page_id
+        )
+        cloned_components.extend(cloned_component_list_for_page)
+        cloned_lizts.extend(cloned_lizt_per_page)
+    db.session.add_all([clone, *cloned_pages, *cloned_components, *cloned_lizts])
     cloned_pages = _fix_cloned_default_pages(cloned_pages)
     db.session.commit()
 
@@ -115,20 +130,28 @@ def _initiate_cloned_components_for_page(
     components_to_clone: list[Component], new_page_id: str = None, new_theme_id: str = None
 ):
     cloned_components = []
+    cloned_lizts = []
     for component_to_clone in components_to_clone:
         cloned_component = _initiate_cloned_component(
             component_to_clone, new_page_id=new_page_id, new_theme_id=None
         )  # TODO how should themes work when cloning?
+        if component_to_clone.list_id:
+            # clone lizt if component has a lizt and map it cloned component
+            cloned_lizt = _initiate_cloned_lizt(component_to_clone.lizt)
+            cloned_component.list_id = cloned_lizt.list_id
+            cloned_lizts.append(cloned_lizt)
         cloned_components.append(cloned_component)
-    return cloned_components
+    return cloned_components, cloned_lizts
 
 
 def clone_single_page(page_id: str, new_form_id=None) -> Page:
     page_to_clone: Page = db.session.query(Page).where(Page.page_id == page_id).one_or_none()
     clone = _initiate_cloned_page(page_to_clone, new_form_id)
 
-    cloned_components = _initiate_cloned_components_for_page(page_to_clone.components, new_page_id=clone.page_id)
-    db.session.add_all([clone, *cloned_components])
+    cloned_components, cloned_lizts = _initiate_cloned_components_for_page(
+        page_to_clone.components, new_page_id=clone.page_id
+    )
+    db.session.add_all([clone, *cloned_components, *cloned_lizts])
     db.session.commit()
 
     return clone
@@ -139,6 +162,12 @@ def clone_single_component(component_id: str, new_page_id=None, new_theme_id=Non
         db.session.query(Component).where(Component.component_id == component_id).one_or_none()
     )
     clone = _initiate_cloned_component(component_to_clone, new_page_id, new_theme_id)
+    if component_to_clone.list_id:
+        # clone lizt if component has a lizt and map it cloned component
+        cloned_lizt = _initiate_cloned_lizt(component_to_clone.lizt)
+        clone.list_id = cloned_lizt.list_id
+        db.session.add(cloned_lizt)
+        db.session.commit()
 
     db.session.add(clone)
     db.session.commit()
--- a/tests/unit/test_clone.py
+++ b/tests/unit/test_clone.py
@@ -2,13 +2,14 @@ from uuid import uuid4
 
 import pytest
 
-from app.db.models import Component, ComponentType, Page
+from app.db.models import Component, ComponentType, Lizt, Page
 from app.db.models.application_config import Form, Section
 from app.db.models.round import Round
 from app.db.queries.clone import (
     _fix_cloned_default_pages,
     _initiate_cloned_component,
     _initiate_cloned_form,
+    _initiate_cloned_lizt,
     _initiate_cloned_page,
     _initiate_cloned_section,
     clone_multiple_components,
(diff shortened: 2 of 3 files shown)


wjrm500:
Chatted to @NarenderRajuB this morning - he made a convincing case for this change being executed via an Alembic migration rather than a one-off script, so I humbly retract that part of my review. Once we've removed the duplicate `list_id` in the `Component` SQLAlchemy model, I can re-review this PR

app/db/migrations/versions/~2025_03_07_0017-6f5560b34904_clone_lizt_for_each_component.py
```diff
@@ -0,0 +1,79 @@
...
+        if old_lizt:
+            # Clone the Lizt
+            new_lizt = Lizt(
+                list_id=uuid.uuid4(),
+                name=old_lizt.name,
+                title=old_lizt.title,
+                type=old_lizt.type,
+                items=old_lizt.items,
+                is_template=False,
+            )
```
nuwan-samarasinghe, line 48:
```
session.expunge(original_lizt)
make_transient(original_lizt)
original_lizt.list_id = uuid.uuid4()
```

@wjrm500 @NarenderRajuB let me know your thoughts about this 

so here we dont need to worry about the attributes that we have if we use above

also we can use this for cloning

NarenderRajuB: > ```
> session.expunge(original_lizt)
> make_transient(original_lizt)
> original_lizt.list_id = uuid.uuid4()
> ```
> 
> @wjrm500 @NarenderRajuB let me know your thoughts about this
> 
> so here we dont need to worry about the attributes that we have if we use above
> 
> also we can use this for cloning

here @nuwan-samarasinghe  this will remove the original_lizt from session, we may need to add it back and new row , which will have new id and may need to change the existing mapping

app/db/migrations/versions/~2025_03_07_0017-6f5560b34904_clone_lizt_for_each_component.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = "6f5560b34904"
+down_revision = "7b151261e00d"
+branch_labels = None
+depends_on = None
+
+
+dSession = sessionmaker()
```
wjrm500, line 26: This can be removed

remove_dummy_data.py
wjrm500, line 1: Let's put this in a `scripts/` directory

wjrm500 (changes requested):
Hi @NarenderRajuB, I've just taken another look at this and, taking into consideration the discussion we had today around data migrations versus one-off repair scripts, I stand by my original review - I feel that we should not be doing a data migration here.

The reasons for this are:
- Migrations should be for moving from a known correct state to a known correct state, in line with code changes to data models
- _This_ is a schema-agnostic data repair job - previously we were erroneously re-using a single list across multiple components during cloning, causing weird behaviour
- I don't see why we'd ever want to revert to a known erroneous behaviour
- We are also losing a crucial piece of information on upgrade - the original list ID - that then has to be inferred via non-unique fields name and title on downgrade - which is flaky
- Therefore to me a migration just doesn't make sense here
- I understand that migrations are run automatically on deployment which saves developer time and provides consistency across environments, but ultimately it's pretty trivial for a developer to simply go through the environments one-by-one and run the repair script on a one-off basis

**Please follow the approach I outlined in my comment [here](https://mhclgdigital.atlassian.net/browse/FS-5036?focusedCommentId=655220) on the 3rd March.** Let me know if there are any genuine blockers as this ticket has been idling for a while and is preventing us from clearing out dummy data from the UAT environment. Thanks!

NarenderRajuB:
> * The current downgrade behaviour is very flaky - when we upgrade, we're creating a bunch of cloned lists with the same name and the same title. Then when we downgrade, we search for the original list by querying name and title - meaning we could just be grabbing a random cloned list (there is no guarantee `first()` will return the original list). The downgrade logic also relies on keeping orphaned lists in the database just in case we ever want to revert, which is super confusing

here when we downgrade we search for lizt with is_template=True which is actual list

NarenderRajuB:
> > here when we downgrade we search for lizt with is_template=True which is actual list
> 
> This filter doesn't appear to be in the current downgrade method?

missed the changes i have pushed them now.. meanwhile i will work on script for the same.

app/db/queries/clone.py
```diff
@@ -100,35 +103,56 @@ def clone_single_form(form_id: str, new_section_id=None, section_index=0) -> For
...
 ):
     cloned_components = []
+    cloned_lizts = []
     for component_to_clone in components_to_clone:
         cloned_component = _initiate_cloned_component(
             component_to_clone, new_page_id=new_page_id, new_theme_id=None
         )  # TODO how should themes work when cloning?
+        if component_to_clone.list_id:
+            # clone lizt if component has a lizt and map it cloned component
+            cloned_lizt = _initiate_cloned_lizt(component_to_clone.list_id)
```
wjrm500, line 141: Better to pass the `Lizt` object itself into the `_initiate_cloned_lizt` function (i.e., by `_initiate_cloned_lizt(component_to_clone.lizt)`) for consistency with other `_initiate_cloned...` functions and to eliminate the need for this line `to_clone: Lizt = db.session.query(Lizt).where(Lizt.list_id == lizt_id).one_or_none()` in `_initiate_cloned_lizt`

app/db/queries/clone.py
```diff
@@ -100,35 +103,56 @@ def clone_single_form(form_id: str, new_section_id=None, section_index=0) -> For
...
+        cloned_components.extend(cloned_component_list_for_page)
+        cloned_lizts.extend(cloned_lizt_per_page)
+    db.session.add_all([clone, *cloned_pages, *cloned_components, *cloned_lizts])
     cloned_pages = _fix_cloned_default_pages(cloned_pages)
     db.session.commit()
 
     return clone
 
 
+def _initiate_cloned_lizt(lizt_id: str) -> Lizt:
```
wjrm500, line 122: Could we move this function up the module so that it sits just underneath the other `_initiate_cloned...` functions?

clone_lizts_for_cloned_components.py
```diff
@@ -0,0 +1,49 @@
...
+            old_lizt = Lizt.query.filter_by(list_id=component.list_id).first()
+            if old_lizt:
+                # Clone the Lizt
+                new_lizt = Lizt(
+                    list_id=uuid.uuid4(),
+                    name=old_lizt.name,
+                    title=old_lizt.title,
+                    type=old_lizt.type,
+                    items=old_lizt.items,
+                    is_template=old_lizt.is_template,
```
wjrm500, line 24: I think this should be set to `False`?

clone_lizts_for_cloned_components.py
```diff
@@ -0,0 +1,49 @@
...
+                db.session.add(new_lizt)
+                db.session.flush()  # Ensure ID is generated
+
+                # Update the component to point to the new Lizt
+                component.list_id = new_lizt.list_id
+                db.session.add(component)
+
+        db.session.commit()
+        print("✓ Successfully cloned Lizts")
+    except Exception:
```
wjrm500, line 35: By capturing the exception here and doing a print with no detail, we are preventing errors reaching the main exception handler, which is where we actually stringify the error for detail

clone_lizts_for_cloned_components.py
wjrm500, line 1: Please move this and the other script into the `scripts` directory

wjrm500, line 1: Please add a comment to the top of this script explaining briefly what it is doing and why, thanks

remove_dummy_data.py
```diff
@@ -0,0 +1,114 @@
...
+    "Sarah Fund 1",
+    "Water bottle fund",
+]
+
+# Rounds to delete by English name and fund English name and short name
+ROUNDS_TO_DELETE = [
+    {"fund_name": "Madaline Hunter", "round_name": "Round 1"},
+    {"fund_name": "Madaline Hunter", "round_name": "Est et tempora qui q"},
+    {
+        "fund_name": "Community Ownership Fund 2025",
```
wjrm500, line 27: Have we double checked with Craig which of these COF rounds he wants deleted? They look more legitimate than the other dummy rounds here, and there's three near-duplicates, so we should definitely double check here before executing the deletion

remove_dummy_data.py
```diff
@@ -0,0 +1,114 @@
...
+    "Madaline Hunter",
+    "Napton Flower Festival",
+    "Pauline's Fund",
+    "Ruman's Fund 2024/2025",
+    "Sarah Fund 1",
+    "Water bottle fund",
+]
+
+# Rounds to delete by English name and fund English name and short name
+ROUNDS_TO_DELETE = [
```
wjrm500, line 23: You've added `round_short_name` to some rounds but not all, but then in the filter below we have `Round.short_name == round_info["round_short_name"]`, which will break where we have no `round_short_name`

tests/unit/test_clone.py
```diff
@@ -145,6 +146,29 @@ def test_initiate_cloned_component(mock_new_uuid):
 # =====================================================================================================================
 # These functions test the clone_XXX functions and DO use the db
 # =====================================================================================================================
+def test_clone_lizt_for_component(flask_test_client, _db):
```
wjrm500, line 149: Not sure this function name makes sense - the current test implementation is just testing the `Lizt` cloning - no reference to the higher-level `Component` entity here - better name would be simply `test_clone_lizt`

tests/unit/test_clone.py
```diff
@@ -145,6 +146,29 @@ def test_initiate_cloned_component(mock_new_uuid):
...
+    lizt_component: Lizt = Lizt(list_id=uuid4(), items=items, type="string", is_template=True, title="dataset formats")
+
+    old_id = lizt_component.list_id
+
+    _db.session.bulk_save_objects([lizt_component])
+    _db.session.commit()
+
+    assert _db.session.get(Lizt, old_id)
+
+    result = _initiate_cloned_lizt(lizt_component.list_id)
```
wjrm500, line 164: `cloned_lizt` is a better variable name

wjrm500 (changes requested):
Hi @NarenderRajuB, thanks for updating the PR. I've added some comments - only one major issue (the `round_short_name` reference in the dummy data removal script), one moderate issue (not setting to `is_template` to `False` for clones in the cloning script) and then a few more minor issues that should still be addressed. Please also reach out to Craig as mentioned in one comment

The PR would benefit also from including the ticket number in the title, an improved commit message (the current one "Fix lizts cloning and added script for removing dummy data# This is a combination of 2 commits." is a bit weird), and a more thorough description, including proposed next steps

remove_dummy_data.py
```diff
@@ -0,0 +1,114 @@
...
+    "f",
+    "Frog Road Safety Fund",
+    "Jena Allen",
+    "Madaline Hunter",
+    "Napton Flower Festival",
+    "Pauline's Fund",
+    "Ruman's Fund 2024/2025",
+    "Sarah Fund 1",
+    "Water bottle fund",
+]
```
nuwan-samarasinghe, line 20: I prefer if we can give funds and rounds as parameters while executing the script ? @wjrm500 any comments what is the usage of this script other wise do we need to manually update this and run it locally cant we have this as a proper script which allow us to give parameters and delete funds or rounds if needed ?

nuwan-samarasinghe (changes requested):

nuwan-samarasinghe:
@NarenderRajuB looked on your PR following were identified 

1. Unit tests are failing 
2. I think this entire clone logic is overkilling we should simplify it (No need to do now)
3. Also I agree with the comments already here

remove_dummy_data.py
NarenderRajuB: deleted all rounds which were deleted already in UAT

remove_dummy_data.py
NarenderRajuB: I did check with @Craig on this , he did asked me to delete COF fully, but still i am deleting only cloned Application rounds. I am now moving this script out of this PR as i have few questions on why we need this inside script directory this is one of run script

clone_lizts_for_cloned_components.py
NarenderRajuB: Done i will creating another PR for just these two scripts,I am now moving this script out of this PR as i have few questions on why we need this inside script directory this is one of run script

NarenderRajuB: I am now moving this script out of this PR as i have few questions on why we need this inside script directory this is one of run script

clone_lizts_for_cloned_components.py
NarenderRajuB: Done

NarenderRajuB:
> @NarenderRajuB looked on your PR following were identified
> 
> 1. Unit tests are failing
> 2. I think this entire clone logic is overkilling we should simplify it (No need to do now)
> 3. Also I agree with the comments already here

fixed the tests . I have now removed the scripts from this PR. will be creating a new PR

clone_lizts_for_cloned_components.py
NarenderRajuB: PR added https://github.com/communitiesuk/funding-service-design-fund-application-builder/pull/337

clone_lizts_and_remove_dummy_data.py
```diff
@@ -0,0 +1,151 @@
...
+from app.db.models.application_config import Component, Lizt
+from app.db.models.fund import Fund
+from app.db.models.round import Round
+from app.db.queries.fund import delete_selected_fund
+from app.db.queries.round import delete_selected_round
+
+# Funds to delete by English name
+FUNDS_TO_DELETE = [
+    "Crash Test Dummy Fund",
+    "f",
```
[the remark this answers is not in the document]
NarenderRajuB: PR https://github.com/communitiesuk/funding-service-design-fund-application-builder/pull/337 addresses this

remove_dummy_data.py
NarenderRajuB: PR https://github.com/communitiesuk/funding-service-design-fund-application-builder/pull/337 addresses this @nuwan-samarasinghe

app/db/queries/clone.py
wjrm500: Unresolved as this was not actioned - minor thing but please @NarenderRajuB 🙏

wjrm500 (approved):
Looks good @NarenderRajuB, PR description is non-existent but at this point we can ignore that.

One other (very minor) thing - you resolved my comment about moving the `_initiate_cloned_lizt` function further up the `clone.py` module without actioning this / replying - if you could make that adjustment then great.

But approving regardless to unblock 👍

nuwan-samarasinghe (approved):
Looks ok for me

NarenderRajuB:
> _initiate_cloned_lizt

Updated . Please review again

PR #331, opened 2025-03-12
FS-4933 - Add search functionality for grants, applications and templates tables

wjrm500:
### Ticket

[Add Search Functionality for Grants, Applications and Templates tables](https://mhclgdigital.atlassian.net/browse/FS-4933)

### Description

This PR adds search functionality to the Grants, Applications, and Templates table pages, making it easier for users to find specific items in large lists. Each page now has a search bar above the table where the user can enter a search query, firing it off either by pressing the "Enter" key or clicking the "Search" button to the right. The search is case-insensitive and matches any part of the search field. In the case of funds, the search field is the fund name; in the case of rounds, it is the round title; in the case of templates it is the template name. If a search has been made, a new button "Clear search" will be visible to the right of the search bar and button.

### Specific changes

- Added a search bar with "Search" button and "Clear search" option that appears after searching
- Search uses SQL ILIKE for case-insensitive substring matching across relevant name fields
- Search term is preserved when navigating between pages of results
- Comprehensive tests verify search works with prefixes, internal substrings, and case variations

### How to test

1. Navigate to the Grants, Applications, or Templates pages
2. Enter a search term that matches an existing item (try both full and partial matches)
3. Verify that only matching items are displayed
4. Test case insensitivity by entering the same term with different capitalisation
5. Click "Clear search" and verify all items return
6. Try clicking through pagination while a search is active to verify the search term persists

### Screenshots of UI changes

image

image


app/templates/macros/search.html
```diff
@@ -0,0 +1,26 @@
...
+    <div class="govuk-grid-column-full">
+        <form action="" method="GET" novalidate>
+            <div class="govuk-form-group">
+                <label class="govuk-label" for="search">
+                    Search {{ entity_name }}
+                </label>
+                <div class="govuk-grid-row">
+                    <div class="govuk-grid-column-two-thirds govuk-!-padding-right-0">
+                        <input class="govuk-input" id="search" name="search" type="text" value="{{ current_search_term }}">
+                    </div>
```
nuwan-samarasinghe, line 12: We'll use wtforms Inputfield component ?

1. To not to brake consistency and it easy to control validations if required  (may be here we dont need validation)
2. Form handling logic stays in your Python code

but recoment to use wtforms

app/templates/macros/search.html
```diff
@@ -0,0 +1,26 @@
...
+                    <div class="govuk-grid-column-one-third govuk-!-padding-left-0" id="search-bar-right">
```
nuwan-samarasinghe, line 13: I prefer to have the search & relevant controllers same width as the table.. but leve my comments aside :)

app/static/src/styles/fab.css
```diff
@@ -109,3 +109,8 @@
     margin-bottom: 2px;
     margin-left: 15px;
 }
+
+#search-bar-right {
+    display: flex;
+    align-items: center;
+}
```
nuwan-samarasinghe, line 116: `govuk-!-display-flex govuk-!-align-items-center`

shall we check this ? I did not check this if this is not there please ignore

nuwan-samarasinghe (changes requested):

app/static/src/styles/fab.css
wjrm500: I tried it! But good thinking

app/templates/macros/search.html
wjrm500: Haha that's one for Mika, I think he replied in the Slack thread around this:

> We chose this width to make it look different from other (metadata) fields that are full width. Also to reduce proximity with other UI.

app/templates/macros/search.html
wjrm500: Fair to raise - off the back of your comment, I had a look at what implementing a WTForms approach would look like here and I don't think the added complexity is justified for a simple search form with no validation, CSRF protection etc.

nuwan-samarasinghe:
I have done a second look around on your PR and following were identified 

1. Unit tests need a fix 
2. May be this is a content design comment because as a user when I tried search on grant, application. it does not clearly mention that you can search stuff from first column data @chaneldiep please take a look on this 
3. I tested the app without the custom css & with that but could not find any change probably @wjrm500 we'll pair and check this together why you need it ?
4. Also in the css I saw `display: flex; align-items: center;` this is repeated with different name it is not a good practice we can introduce a meaningful name & change the places  (this can be a separate PR )
5. Search and clear search is not aligned better if we have both of them aligned
6. This might not an issue since we don't have that much data but if we try to use this search it will scan the full table since we are not using any database indexing.

nuwan-samarasinghe:
My self and @wjrm500 had a chat about the above changes & he has agreed to create a new PR for some of the changes

nuwan-samarasinghe (approved):

wjrm500:
> I have done a second look around on your PR and following were identified
> 
> 1. Unit tests need a fix
> 2. May be this is a content design comment because as a user when I tried search on grant, application. it does not clearly mention that you can search stuff from first column data @chaneldiep please take a look on this
> 3. I tested the app without the custom css & with that but could not find any change probably @wjrm500 we'll pair and check this together why you need it ?
> 4. Also in the css I saw `display: flex; align-items: center;` this is repeated with different name it is not a good practice we can introduce a meaningful name & change the places  (this can be a separate PR )
> 5. Search and clear search is not aligned better if we have both of them aligned
> 6. This might not an issue since we don't have that much data but if we try to use this search it will scan the full table since we are not using any database indexing.

Thanks @nuwan-samarasinghe 

1. Unit test has been fixed
2. Agreed this is strange, let's flag to content design
3. Change is only visible when "Clear search" is visible, it's a minor alignment adjustment demoed on call today
4. Agreed this seems duplicative, good spot. As the duplicated CSS touches a few different places, best to dump into a separate PR
5. Think this issue may have been related to your testing of point 3
6. Another good point, performance won't be a practical concern for the foreseeable and I'm reluctant to make a targeted optimisation at this stage as I have a strong feeling we're not going to be searching on "Apply for {fund_title}" for too long once the weirdness is flagged

PR #333, opened 2025-03-12
Fix failing export

wjrm500:
Fix failing export


nuwan-samarasinghe (approved):

wjrm500:
Tested this by...

- Restoring UAT data dump locally
- Verifying that the exported HTML for the rounds DPIF R3 and COF R1 is identical between `main` branch and this one
- Verifying that export succeeds for the broken round Communities for Afghans R1

Moving forward, we should take a closer look at improving the robustness of `metadata_utils.py`

PR #334, opened 2025-03-12
Fix the flash message once created the application

nuwan-samarasinghe:
### Change description
Fix flash message once created the application


wjrm500 (approved):
Not 100% sure this is a bug - more like 90% - so happy to proceed on that basis, given the cost is minimal

PR #339, opened 2025-03-17
adding an environment variable for Flask-DebugToolabr activation in development mode

nuwan-samarasinghe:
### Change description
As a developer if I wanted to disable the flask debugtool in the local environment previously there was not a way to set it to False by using environment variable and here we have added a option to disable it when ever needed and this is a environment variable and need to give 

`export DEBUG_TB_ENABLED=False` if you dont need debugtool in local

### How to test
_If manual testing is needed, give suggested testing steps_


config/envs/development.py
```diff
@@ -33,8 +33,8 @@ class DevelopmentConfig(Config):
     DEBUG_USER_ACCOUNT_ID = "00000000-0000-0000-0000-000000000000"
 
     # Flask-DebugToolabr
-    DEBUG_TB_ENABLED = True
-    DEBUG_TB_INTERCEPT_REDIRECTS = False
+    DEBUG_TB_ENABLED = getenv("DEBUG_TB_ENABLED", True)
+    DEBUG_TB_INTERCEPT_REDIRECTS = getenv("DEBUG_TB_INTERCEPT_REDIRECTS", False)
```
samuelhwilliams, line 37: getenv will return a string - can we convert to a bool please? 🙏

nuwan-samarasinghe: first I thought this changes are working and it seems these `DEBUG_TB_ENABLED` need to explicitly mention False or True (0 or 1) if we wanted to stop or display the toolbar pop up, sending a environment variable is not working bit of a wired situation so taking this PR down since this is not getting any value for larger investigation

wjrm500:
I've updated this PR to do the Boolean-ifying and can confirm is working as expected locally. You need to make a change in the `.env` file of the Docker Runner repo before re-building the FAB container @nuwan-samarasinghe, this might be where you went wrong? See this PR in Docker Runner: https://github.com/communitiesuk/funding-service-design-docker-runner/pull/136

samuelhwilliams (approved):

PR #343, opened 2025-03-17
Adding a fix for the CICD flow chain

nuwan-samarasinghe:
Fix dependency chains in deployment workflow. Specifically, we have made unit tests a dependency of the build step and leveraged dependency hierarchies to simplify configuration.


wjrm500 (approved):
Paired with @nuwan-samarasinghe on this

PR #344, opened 2025-03-17
Adding simple names to all GitHub workflow jobs for readability

wjrm500:
### Description

We are missing job names on our GitHub workflow jobs, meaning that we are forced to use whatever label we've assigned to the job in the code e.g., `dev_deploy`.

This PR adds human-readable names to all jobs.

### Screenshot of UI changes

image


nuwan-samarasinghe (approved):
Nice naming 

I saw required checks are just waiting probably need to change into these names

PR #350, opened 2025-03-18
Fix - re-adding necessary GitHub workflow dependencies

wjrm500:
An attempt was made to simplify the dependency chain in `copilot_deploy.yml` recently but it missed the fact that jobs in this workflow rely on taking outputs directly from their dependencies e.g., via `needs.setup.outputs.jobs_to_run`.

This caused breakages, so these dependencies should be restored.


samuelhwilliams:
✅  nice commit msg too

PR #352, opened 2025-03-19
Run e2e test against pr to verify the changes

nuwan-samarasinghe:
**Ticket:  https://mhclgdigital.atlassian.net/browse/FS-5020**

### Change description

When user creates a PR it should run the e2e locally inside GitHub actions and verify the changes are not going to brake any of the user journeys 

**Why we did not used the packito image ?**

when we install that in the AWS environment then that environment has the access level needed to run mkdir, file move etc.. but when we start the container inside the Github it gives a permission denied error

also there is no much difference the way that we star the app since we use the developed code & sole purpose to check the code works according to the user flows

**What changes that have implemented ?**

starting the Postgres sql in the GitHub workflow and apply the db changes finally start app & run against local environment

### DIFF
--- a/.github/workflows/copilot_deploy.yml
+++ b/.github/workflows/copilot_deploy.yml
@@ -33,9 +33,52 @@ jobs:
       postgres_unit_testing: true
       db_name: fab_store_test
 
+  e2e:
+    if: github.event_name == 'push'
+    name: Run E2E tests
+    permissions:
+      id-token: write
+      contents: read
+    runs-on: ubuntu-latest
+    container: mcr.microsoft.com/playwright:v1.47.0-noble@sha256:3d41153494e2b12a5a5fa6e26cf1e854c3997d13758754d46e7bf902b5ba09b1
+    services:
+      postgres:
+        image: postgres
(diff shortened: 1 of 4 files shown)


wjrm500 (changes requested):
Great work, this is a complex ticket and you've got a working solution up already. A couple of changes though please...

- We should use the Paketo-built image to run FAB as a container on the Playwright container (containers in containers :octocat:), rather than running Flask directly on that container. Why? We already do the Paketo build step, and using the image gives us more reassurance
- We should take this chance to break apart the copilot_deploy.yml workflow into *what runs in PRs* and *what runs on merge to main* and then incorporate our new PR-scoped E2E tests in the former
- We should make sure that the PR-scoped E2E tests do not run on push to main, as this is duplicative

I'll pop you a message on Slack to discuss further

wjrm500 (approved):
Discussed with @nuwan-samarasinghe about why we don't use a Paketo-built image here and it's fine - but we've separated out the workflows into two files now which feels clean, and everything seems to be working!

PR #363, opened 2025-03-24
FS-5063 - Adding concurrency group to deployment jobs in GitHub workflow to prevent accidental deployment from outdated workflow runs

wjrm500:
### Ticket

[Prevent outdated GitHub workflows running (FAB)](https://mhclgdigital.atlassian.net/browse/FS-5063)

### Description

If a previous deployment hasn't been cancelled, when you try to manually approve a later deployment, you can be silently switched to the previous workflow run, making accidental deployment of outdated workflow runs a significant hazard.

This change puts each deployment job (Dev, Test and UAT) in concurrency groups (e.g., "deploy-dev"), meaning that when they begin to run, existing deployments to the corresponding environment in previous workflow runs will be automatically cancelled, which should in theory prevent the weird behaviour described.

See [concurrency](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#concurrency) section of GitHub Actions docs for reference.

### DIFF
--- a/.github/workflows/copilot_deploy.yml
+++ b/.github/workflows/copilot_deploy.yml
@@ -50,6 +50,9 @@ jobs:
     name: Deploy to Dev
     needs: [ paketo_build ]
     if: ${{ contains(fromJSON(needs.setup.outputs.jobs_to_run), 'dev') }}
+    concurrency:
+      group: deploy-dev
+      cancel-in-progress: false
     uses: communitiesuk/funding-service-design-workflows/.github/workflows/standard-deploy.yml@main
     secrets:
       AWS_ACCOUNT: ${{ secrets.AWS_ACCOUNT }}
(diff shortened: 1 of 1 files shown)


.github/workflows/copilot_deploy.yml
```diff
@@ -100,6 +100,9 @@ jobs:
     needs: [ test_e2e_test, paketo_build, setup ]
     if: ${{ always() && contains(fromJSON(needs.setup.outputs.jobs_to_run), 'uat') && (! contains(needs.*.result, 'failure') ) && (! contains(needs.*.result, 'cancelled') )}}
     name: Deploy to UAT
+    concurrency:
+      group: deploy-uat
+      cancel-in-progress: true
```
MarcUsher, line 105: In other repo workflows where we use concurrency we have `cancel-in-progress` set to `false`, to avoid it potentially trying to cancel an in-progress AWS deploy and leaving the resources in a weird state (in reality I don't think that'd happen as even if the workflow was cancelled the deployment in AWS would continue once the copilot command was executed but I think we've set to `false` just for safety).

MarcUsher (commented):
Noice 🙌 could be worth adding concurrency to `test_deploy` as well just to avoid any strange behaviour if two PRs get merged in quick succession (which would presumably try to kick off a second copilot deploy while the first one is still in progress)?

MarcUsher (approved):

PR #364, opened 2025-03-24
FS-4954 - Simplifying the 'Build application' page when there are no sections in an application

wjrm500:
### Ticket

[Mark application as complete](https://mhclgdigital.atlassian.net/browse/FS-4954)

### Description

The 'Build application' page is much simplified when there are no sections in an application. The following elements are hidden:

- "Status: X" text
- "Download application ZIP file" button
- "View all application questions" button
- "Mark application complete" button

This should have been implemented as part of the original work on 'Mark application as complete' but was overlooked.

### Screenshots of UI changes

image


nuwan-samarasinghe (approved):
Yeah this is a good to have feature did not locally test checked the image

PR #369, opened 2025-03-25
FS-4810 - Putting FAB into production

wjrm500:
### Ticket

[Putting FAB in production environment](https://mhclgdigital.atlassian.net/browse/FS-4810)

### Description

We are deploying FAB into production. FAB has been used by users since January, but the top level env has always been UAT. Once we get out to production, we need to restore a dump of the UAT database into production and notify users of the switchover. 

### Notes

Before converting this PR out of draft, we need to..

- [x] - Remove `deploy-to-prod` from the allowed branches for the deployment workflow
- [x] - Re-add the "determine jobs" condition to the prod deploy step `if`
- [x] - Set `run_db_migrations` to `true` for the prod deploy step
- [x] - Ensure that prod FAB points to UAT Form Runner for preview links (Form Runner in prod has "PREVIEW_MODE" disabled to prevent unauthorised publishing of JSON forms to the Redis cache - we need to come up with a better long-term solution)
- [x] - Change Aurora engine from 14.4 to 14.9 in CloudFormation add-on

Before deploying to UAT, we should ensure we take a dump of the UAT database. This is because we will be deploying out a CloudFormation stack change to this env for the first time. We have already tested this stack change on Dev and there were no issues, but good to take precautions and we need to dump UAT regardless.

### DIFF
--- a/.github/workflows/copilot_deploy.yml
+++ b/.github/workflows/copilot_deploy.yml
@@ -18,7 +18,7 @@ on:
         type: boolean
         description: Run e2e tests after deployment
   push:
-    branches: [ main, deploy-to-prod ]
+    branches: [ main ]
 
 jobs:
   setup:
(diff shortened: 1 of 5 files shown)


gidsg:
Am thinking this may not work as intended, as this internal host will point to the production form runner: https://github.com/communitiesuk/funding-service-design-fund-application-builder/blob/84475e7ed5f7600f628c475d5174419df57e93d9/copilot/fsd-fund-application-builder/manifest.yml#L50 

Which looks like it is used here: https://github.com/communitiesuk/funding-service-design-fund-application-builder/blob/84475e7ed5f7600f628c475d5174419df57e93d9/config/envs/default.py#L21 

And then hits the `publish` endpoint here:
https://github.com/communitiesuk/funding-service-design-fund-application-builder/blob/7226a83d3c38587034c75fcec15a54fb3147af86/app/blueprints/index/routes.py#L53

So I think this may not function as intended?

nuwan-samarasinghe (approved):
Looks nice

PR #379, opened 2025-03-31
Allow manual deployments to prod via GitHub workflow dispatch

wjrm500:
Currently we can manually deploy branches to each of the Dev, Test and UAT AWS environments via the Actions tab in the GitHub UI, by clicking Deploy to AWS -> Run workflow. The prod environment was recently introduced to FAB, and we should add that as a manually deployable environment. This is in-keeping with the approach taken in the Pre-Award repo.

### DIFF
--- a/.github/workflows/copilot_deploy.yml
+++ b/.github/workflows/copilot_deploy.yml
@@ -12,6 +12,7 @@ on:
           - dev
           - test
           - uat
+          - prod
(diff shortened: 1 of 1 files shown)


wjrm500:
As @MarcUsher points out, we don't actually want this at all. Marc, do you want to do the honours and close this PR without comment, like the glory days?

PR #384, opened 2025-04-07
FLS-1194 - Replace internal user check with allowed domains check

wjrm500:
### Ticket

[FAB - give external users pre-prod access for accessibility audit](https://mhclgdigital.atlassian.net/browse/FLS-1194)

### Description

This change replaces the check on internal user with a FAB-specific check on allowed domains, such that we can provide access to FAB not only to users with communities.gov.uk email addresses, but also email addresses with other domains, configured via AWS secrets to allow for environment-level configuration. We are specifically doing this now because we have an accessibility audit coming up and need to ensure assessors can access.


app/auth.py
```diff
@@ -0,0 +1,33 @@
...
+        fab_domains_str = current_app.config.get("ALLOWED_DOMAINS", "")
+        fab_domains = [f"@{domain.strip()}" for domain in fab_domains_str.split(",") if domain.strip()]
+
+        # Combine all allowed domains
+        all_allowed_domains = internal_domains + tuple(fab_domains)
+
+        authenticated = g.is_authenticated
+        is_allowed_domain = authenticated and any(g.user.email.endswith(domain) for domain in all_allowed_domains)
+
+        if not authenticated or is_allowed_domain:
```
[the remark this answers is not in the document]
wjrm500: The new `check_allowed_domains` drew heavily from the existing `check_internal_user` in `fsd_utils`, which was unchanged when ported from Post-Award (see [here](https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/864/files#diff-7b69651b703ae16f38c6fd626cc9515d817dd1fb24a7ef13b5b223c24e0de852)). This comment flags up the fact that the code is a bit nonsensical - it passes if the user is not authenticated. Seems more logical to remove the authenticated bit from this function and rely on login_required for that, and here simply check email domain is in allowed domains.

app/auth.py
```diff
@@ -0,0 +1,25 @@
...
+def check_allowed_domains(func):
+    """
+    Checks if the authenticated user's email domain is allowed.
+    Combines the internal domains from fsd_utils with additional domains from ALLOWED_DOMAINS.
+    """
+
+    @wraps(func)
+    def decorated(*args, **kwargs):
+        internal_domains = tuple(k.value for k in InternalDomain)
+        fab_domains_str = current_app.config.get("ALLOWED_DOMAINS", "")
```
gidsg, line 16: Could we hard code the existing domains instead, and then merge it with the env var if populated?

wjrm500: We are basically doing that here I think? See `all_allowed_domains = internal_domains + tuple(fab_domains)` below

gidsg (approved):

PR #388, opened 2025-04-15
FLS-1128 - Fixing flaky E2E tests

wjrm500:
### Ticket

[Fix flaky FAB E2E tests](https://mhclgdigital.atlassian.net/browse/FLS-1128)

### Description

The fundamental issue was that when we were trying to select an application on the "Select application" page, we were comparing apples (the round name) with oranges (the round ID), which meant the match would _always_ fail and we'd _always_ default to selecting a completely random application (!), which caused unpredictable downstream effects. For example, if the randomly selected application was marked as "Complete", then the next page would be the more limited "View application" page instead of "Build application". This led to the subsequent check for the "Build application" text to fail, which was the cause of every observed breakage. And if we'd fixed that particular thing (e.g., by checking for either "Build application" OR "View application" text) then we'd have just broken down at the next step when we start testing functionality that is _only_ available in "build mode" (e.g., moving sections up and down).

The solution? Grab the UUID of the round from the application success banner on the dashboard page after the application has been saved, allowing us to select an application from the dropdown by its round ID value. This means we _always_ end up selecting the application that we created in the E2E test rather than a random one, and this means that the application will _always_ be uncompleted (in "build mode"), ready for further testing.

It wasn't clear whether grant selection was similarly messed up, but either way it made sense to leverage the existing `grant_id` metadatum to simplify that logic as well.


hamzabinkhalid (approved):
Looks good 💫

PR #390, opened 2025-04-22
FLS-1235: Add e2e check back for UAT and PROD deploy

hamzabinkhalid:
Ticket: https://mhclgdigital.atlassian.net/browse/FLS-1235

### Change description
- E2E check was disabled for UAT and PROD deploys because they were flaky, but now that they are fixed we should add the checks back in

- [X] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### DIFF
--- a/.github/workflows/copilot_deploy.yml
+++ b/.github/workflows/copilot_deploy.yml
@@ -104,7 +104,7 @@ jobs:
     secrets: inherit # pragma: allowlist secret
 
   uat_deploy:
-    needs: [ test_deploy, paketo_build, setup ]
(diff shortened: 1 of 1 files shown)


wjrm500 (approved):
Looks good, can see this is just reverting the changes from [this PR](https://github.com/communitiesuk/funding-service-design-fund-application-builder/pull/380/files) so can't argue. It does have the side effect that UAT deployment will go back to being optional rather than mandated when getting code through to production, but this doesn't seem obviously negative

PR #396, opened 2025-05-08
Start to retire the UAT environment

samuelhwilliams:
Ticket: https://mhclgdigital.atlassian.net/browse/FSPT-342

https://github.com/communitiesuk/funding-service-requests-for-comments/discussions/17

As per the conclusion of the above RFC, we're moving away from having three pipelined environments to two (test/prod) and a free-for-all pre-merge env (dev).

This patch removes all concepts of the UAT env from the app, but won't tear down the app instances themselves yet. We'll do this manually later when all of our services have been updated and the UAT env is disconnected from all pipelines/etc.


samuelhwilliams:
Pausing pending some more discussion around whether UAT is still in active use (eg for accessibility audit)

PR #405, opened 2025-06-26
FLS-1349: Remove column and add a migration script

nuwan-samarasinghe:
**Ticket: https://mhclgdigital.atlassian.net/browse/FLS-1349**

### Change description

1. Remove children column from the Component db entity.
2. Updating each place of logic to not to populate that data or change them to use the new `children_components` which is a Component type db entity.
3. Delete the data migration script that used to populate child components into Db entity Component


app/db/migrations/versions/~2025_06_26_1547-60c0a648f835_remove_child_column_from_component.py
```diff
@@ -0,0 +1,45 @@
+"""Remove child column from component
```
wjrm500, line 1: Very minor thing but should be `children` not `child`

app/db/migrations/versions/~2025_06_26_1547-60c0a648f835_remove_child_column_from_component.py
wjrm500, line 1: The file name is very long - have we tried running the migrations locally to double check this works?

wjrm500: Have just run migrations locally with `docker compose exec fab uv run flask db upgrade` and can confirm works 👍

app/db/migrations/versions/~2025_06_26_1547-60c0a648f835_remove_child_column_from_component.py
nuwan-samarasinghe: Fixed

PR #421, opened 2025-07-24
FLS-1418 - Ensures clicks on date errors in error summary correctly focus the first field (Day) of multi-part date fields

wjrm500:
_WIP_


nuwan-samarasinghe (approved):
Nice change and hope that is the expected change that DAC is asking for

PR #422, opened 2025-07-24
FLS-1408 - Fix start page logic

wjrm500:
### 🎫 Ticket

[Ensure errors are handled within the same page with summaries.](https://mhclgdigital.atlassian.net/browse/FLS-1408)

### 🎑 Background

This ticket highlighted an issue whereby when we run the DPIF "Name your application" form, and click "Continue" without giving a name, Form Runner exits the scene with a nicely-unhandled 500. As is often the case, FAB fell quickly under suspicion. Basically the process of importing and exporting a perfectly good form into and out of FAB has broken it - FAB nefariously added a controller attribute `start.js` to the first page of the form, meaning that Form Runner no longer recognises that page as a page with user inputs, and error handling functionality is broken. FAB, you swine! We can blame this lovely PR for the bug: https://github.com/communitiesuk/funding-service-design-fund-application-builder/pull/60.

### 🌀 Changes

- **Prevent start page controller being assigned during form import**

    Currently FAB assigns the 'controller' attribute 'start.js' to any page in a form whose path matches the form's start page path. Having this attribute triggers the Form Runner to use the StartPageController to handle the page, which means we lose things like error handling as start pages are not expected by the Runner to contain input components, rather introductory spiel only. And if the unsuspecting content designer fails to include a start page in a form, then because the first page in a form will be automatically assigned as the start page by the Designer, then on import FAB will add the 'controller' attribute and destroy error handling for that page.

   The automatic addition of the 'controller' attribute during form import by FAB seems to serve no purpose whatsoever, especially because during export we create a new start page automatically if it's missing. It looks like a bug that's been kicking around at the heart of FAB for far too long, and now it's time to squash it.

- **Being more explicit about which controller path identifies a start page during form export**

   Form Designer adds the controller './pages/start.js' to start pages, not 'start.js'. The current code technically works because 'start.js' is a substring of './pages/start.js', but we should use the actual controller path.

- **Don't show 'Check your answers' on auto-generated start pages**

   This is relevant now because previously, it was not possible for auto-generation of start pages to actually get hit. This was because if no start page existed on form import, FAB would add the 'controller' attribute 'start.js' to the first page. Then when we export the same form, FAB would always identify the start page by this attribute, and the auto-generation of start page would be skipped.

   Now however, we have removed the automatic assignment of the 'controller' attribute during form import. This is because it is a straight-up bug that basically removes the error-handling capability of the Form Runner on the first pages of forms if the content designer forgets to add a start page to the form - so not great stuff. But the unlocking of start page auto-generation means we need to ensure it is fit for purpose. This change removes the redundant 'Check your answers' from the list of questions we tell the user we're about to ask them on the start page.

### DIFF
--- a/app/export_config/generate_form.py
+++ b/app/export_config/generate_form.py
@@ -8,6 +8,8 @@ from app.db.queries.application import get_list_by_id
 from app.export_config.helpers import human_to_kebab_case
 from app.shared.data_classes import ConditionValue, FormSection
 
+SUMMARY_JS_PATH = "./pages/summary.js"
+
 BASIC_FORM_STRUCTURE = {
     "startPage": None,
     "pages": [],
@@ -33,7 +35,7 @@ SUMMARY_PAGE = {
     "components": [],
     "next": [],
     "section": "uLwBuz",
-    "controller": "./pages/summary.js",
+    "controller": SUMMARY_JS_PATH,
 }
 
 
@@ -149,7 +151,7 @@ def build_navigation(partial_form_json: dict, form: Form) -> dict:
     partial_form_json["conditions"] = build_conditions(form.conditions) if form.conditions else []
 
     for page in form.pages:
-        if page.controller and page.controller.endswith("summary.js"):
+        if page.controller and page.controller.endswith(SUMMARY_JS_PATH):
             continue
 
         this_page_in_results = _get_page_result(partial_form_json, page)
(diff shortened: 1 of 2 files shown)


app/export_config/generate_form.py
```diff
@@ -149,7 +152,7 @@ def build_navigation(partial_form_json: dict, form: Form) -> dict:
     partial_form_json["conditions"] = build_conditions(form.conditions) if form.conditions else []
 
     for page in form.pages:
-        if page.controller and page.controller.endswith("summary.js"):
+        if page.controller and page.controller.endswith(SUMMARY_JS_PATH):
```
nuwan-samarasinghe, line 155: I think better to check the existing data it might not have `./pages/summary.js`

nuwan-samarasinghe (approved):
🥇

PR #423, opened 2025-07-28
FLS-1420 - Don't preselect radios

wjrm500:
### 🎫 Ticket

[[FAB] Avoid pre-selecting radio buttons to prevent confusion for users with memory impairments.](https://mhclgdigital.atlassian.net/browse/FLS-1420)

### 🌍 Background

The recent accessibility audit highlighted that there are pages in FAB where radio buttons are pre-selected on initial load. The first is the "Create grant" page, where the "Is this grant available in Welsh?" radio field defaults to "No". And then on the "Create application" page, there are several radio fields that also default to "No".

### ♻️ Changes

- Do not preselect Welsh radio field on 'Create grant' page
- Do not preselect radio buttons on 'Create round' page
- Removing redundant fields from form class used in 'Create round' page
- Fixing E2E tests by checking required radio fields during application creation


nuwan-samarasinghe (approved):
Nice change 🥇

PR #427, opened 2025-08-04
FLS-1444 - Fixing default Form Runner publish URL

wjrm500:
Need to use HTTPS because Runner expects HTTPS. And need to use external rather than internal hostname to avoid hostname mismatch.

When FAB tries to publish forms to Form Runner, it was using the HTTP protocol and internal service name (`http://form-runner:3009/publish`). However, Form Runner is configured to use HTTPS with SSL certificates issued for the external domain `form-runner.communities.gov.localhost`. 

This caused two issues:
1. Protocol mismatch: HTTP requests to an HTTPS service resulted in SSL handshake errors
2. Hostname mismatch: SSL certificate verification failed because the certificate is valid for `form-runner.communities.gov.localhost`, not the internal service name `form-runner`

This change aligns FAB's configuration with Pre-Award, which already uses the correct HTTPS URL with the external hostname for Form Runner communication.

### DIFF
--- a/config/envs/default.py
+++ b/config/envs/default.py
@@ -18,7 +18,9 @@ class DefaultConfig(object):
         "FUND_APPLICATION_BUILDER_HOST", "https://fund-application-builder.communities.gov.localhost:3011"
     )
     FAB_SAVE_PER_PAGE = getenv("FAB_SAVE_PER_PAGE", "dev/save")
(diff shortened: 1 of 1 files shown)


nuwan-samarasinghe (approved):
Nice fix

PR #435, opened 2025-08-12
FLS-1469 - Script to create and store whole form JSON for existing forms

wjrm500:
### 🎫 Ticket

[Script to create and store whole form JSON for existing forms](https://mhclgdigital.atlassian.net/browse/FLS-1469)

### 🌍 Background

Existing forms in all databases across all environments have only `NULL` values for the `form_json` field. They haven’t needed this, because historically we’ve generated JSON on the fly from the various form entities in our relational schema, in any given circumstance where form JSON is required (HTML generation, Form Runner preview, ZIP file export).

We should populate this field for all forms and make it non-nullable. It is essential we do this before we start moving existing functionality away from using the relational forms storage, towards using whole JSON forms storage, otherwise we will suffer a functional regression.

We can achieve this through a script that uses the existing form JSON generation logic to generate form JSON for each form, and then dump that JSON into the `form_json` column, for all forms without a value in that column. We would run this script across all FAB environments using `copilot svc exec`.

### 🚧 Testing

- Have tested successfully locally with `FLASK_ENV=development uv run python -m scripts.populate_form_json`, observing that all forms were updated

### 🔗 Dependencies

- We need to merge this PR first - https://github.com/communitiesuk/funding-service-design-fund-application-builder/pull/434 - this is because if we merge _this_ one first, run the script, and then some user creates a new form, then we'll not have the `form_json` for it

### ⏩ Next steps

- Run this script in all FAB environments
- Make `form_json` column non-nullable (separate PR)

### DIFF
--- a/scripts/populate_form_json.py
+++ b/scripts/populate_form_json.py
@@ -1,0 +1,105 @@
+"""
+Script to populate the form_json column for existing forms.
+
+This script generates form JSON for all forms that currently have NULL values
+in the form_json column, using the existing form JSON generation logic.
+"""
+
+import sys
+
+from flask import current_app
+from sqlalchemy import text
+
+from app.create_app import create_app
+from app.db import db
+from app.db.models import Form
+from app.export_config.generate_form import build_form_json
+
+
+def populate_form_json():
+    """
+    Populate the form_json column for all forms with NULL values.
+    """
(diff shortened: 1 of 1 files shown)


israr-ulhaq (approved):
The script runs as expected in my local env. 👍

PR #438, opened 2025-08-13
Start using whole form JSON for form preview

israr-ulhaq:
### Ticket

[Start using whole form JSON for form preview 
](https://mhclgdigital.atlassian.net/browse/FLS-1464)

### Change description
  
When we preview a form in the Form Runner (see the preview_form endpoint in app.blueprints.index.routes), we re-build the form JSON at runtime from the set of relational entities we decomposed the JSON file down into at import. Put another way, we call the build_form_json function. 

Now we're directly using form_json as we are storing form_josn in the Form table. Instead of build_form_json now it's form.form_json to make it simple.


wjrm500 (approved):
Wunderbar! ⭐

PR #439, opened 2025-08-13
FLS-1466 - Start using whole form JSON for forms export

wjrm500:
### 🎫 Ticket

[Start using whole form JSON for forms export](https://mhclgdigital.atlassian.net/browse/FLS-1466)

### 🌍 Background

As part of the export process in FAB, whereby we generate a ZIP file full of Pre-Award and Form Runner config, we re-build all of the funding round’s form JSONs at runtime from the sets of relational entities we decomposed JSON files down into at import (see `generate_form_jsons_for_round` in `app.export_config.generate_fund_round_form_jsons`). Put another way, we call the `build_form_json` function.

We are moving away from relational form entities in FAB. Now we are storing whole form JSON directly, we can simply replace the `build_form_json` call with direct access to the `Form` model’s `form_json` attribute.

### 🚧 Testing

- All relevant tests have been updated
- To test yourself...
   - Run FAB locally (run `make pre up` in the https://github.com/communitiesuk/funding-service-design-docker-runner repo)
   - Upload a form created in the Form Designer at https://fund-application-builder.communities.gov.localhost:3011/templates/create (you can find such a form in the `fsd_config` directory of the https://github.com/communitiesuk/digital-form-builder-adapter repo)
   - Create an application (you'll need to create a grant first)
   - Click "Build application" on the new application, add a section to the application, click "Edit section" and add the form that you uploaded to the section
   - Click "Download application ZIP file"
   - Confirm that the JSON in the `form_runner` directory of the ZIP file is _identical_ to the JSON file that you uploaded

### DIFF
--- a/app/export_config/generate_fund_round_form_jsons.py
+++ b/app/export_config/generate_fund_round_form_jsons.py
@@ -3,9 +3,7 @@ import json
 from flask import current_app
 from jsonschema import ValidationError
 
-from app.db.queries.fund import get_fund_by_id
 from app.db.queries.round import get_round_by_id
-from app.export_config.generate_form import build_form_json
 from app.export_config.helpers import write_config
 from app.shared.json_validation import validate_form_json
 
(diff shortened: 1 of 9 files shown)


hamzabinkhalid:
Only the name of the form was updated while comparing the jsons, which is expected, looks good 👍 

Screenshot 2025-08-14 at 12 24 10

PR #440, opened 2025-08-14
FLS:1465 use whole json in questions endpoints

hamzabinkhalid:
Ticket: https://mhclgdigital.atlassian.net/browse/FLS-1465

### Change description
- Use whole json instead of rebuilding the form in /all-questions and /questions endpoints
- Test was failing because generate_metadata() in metadat_utils.py (app/all_questions) expects a populated form json
- Added minimal form json structure for form json and validated the page title from form json in test

### How to test
- User should be able to see the all questions pages successfully

### DIFF
--- a/app/blueprints/application/routes.py
+++ b/app/blueprints/application/routes.py
@@ -146,7 +146,7 @@ def view_all_questions(round_id):
     sections_in_round = round.sections
     section_data = []
     for section in sections_in_round:
-        forms = [{"name": form.runner_publish_name, "form_data": build_form_json(form)} for form in section.forms]
+        forms = [{"name": form.runner_publish_name, "form_data": form.form_json} for form in section.forms]
         section_data.append({"section_title": section.name_in_apply_json["en"], "forms": forms})
 
     print_data = generate_print_data_for_sections(
--- a/app/blueprints/template/routes.py
+++ b/app/blueprints/template/routes.py
@@ -16,7 +16,6 @@ from app.db.queries.application import (
     update_form,
 )
 from app.export_config.generate_all_questions import generate_html
-from app.export_config.generate_form import build_form_json
 from app.shared.forms import DeleteConfirmationForm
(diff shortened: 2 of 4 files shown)


wjrm500 (approved):
Thanks @hamzabinkhalid, beautiful how such a simple change `build_form_json(form)` to `form.form_json` abstracts such a fundamental shift in how forms work in FAB ❤️ 

I feel that we do need some more test coverage around HTML generation - what we currently have is pretty inadequate:

- No tests for the `view_all_questions` route
- Only one test for the `template_questions` route that uses a form with no pages at all
- No tests for `generate_print_data_for_sections` or `generate_html` which seem to be key functions in the HTML generation process

But... I'm tempted to say let's not worry about that in this PR, given (A) we don't already have tests and we're not adding _new_ functionality, so we're not technically making the problem any worse; and (B) we may delegate HTML generation to Pre-Award in the not-too-distant future (I still need to create a ticket for this, but the idea is that it'll follow on from the work Israr is doing on building a new endpoint in Pre-Award to receive fund round config)

I have tested manually, using the following steps:

- Both locally (with this branch checked out) and on FAB in Test environment, I uploaded four templates - "Applicant information" and "Asset information" from COF, "Projects" from PFN, and "Dataset information" from DPIF
- Created an identical grant and application both locally and on Test
- Built the applications in an identical way both locally and on Test, adding two sections, the first with "Applicant information" and "Asset information", and the second containing "Projects" and "Dataset information"
- Clicked "View all application questions" both locally and on Test
- Compared the HTML

I found there were a couple of minor discrepancies, specifically...

- "Risk of closure" appears underneath "The asset in community ownership" locally; while on Test, it appears slightly further down, underneath "Public ownership"
- "Terms of your lease" appears underneath "Upload asset valuation or lease agreement" locally; while on Test, it appears slightly further down, underneath "Expected terms of your ownership or lease"

Neither of these strike me as particularly problematic and I am happy to accept them

Approving 👍

PR #441, opened 2025-08-19
FLS-1476 - Use whole form JSON for assessment config export

wjrm500:
### 🎫 Ticket

[Start using whole form JSON for assessment config export](https://mhclgdigital.atlassian.net/browse/FLS-1476)

### 🌍 Background

As part of the export process in FAB, whereby we generate a ZIP file full of Pre-Award and Form Runner config, we generate a minimal assessment config, grouping questions into assessment criteria, subcriteria and themes using a naive mapping from the application config (section → criteria, form → subcriteria, page → theme). This process is tightly coupled to FAB’s relational forms storage model, leveraging ORM-powered syntax.

We are moving away from relational form entities in FAB. Now we are storing whole form JSON directly. We therefore need to rewrite the assessment config generation logic so that it works with whole JSON blobs, in order to avoid a functional regression.

### ♻️ Changes

This change simply replaces any code that relies on soon-to-be-deprecated SQLAlchemy models like Page with code that can simply reference the whole form JSON stored in the form_json column of the form table. In some cases this means leveraging code from load_form_json, which is where we create models from the form JSON, e.g., I can see in load_form_json that Component.runner_component_name is just component.get('name'), therefore we can swap out out component.runner_compnent_name for component.get('name') in our updated assessment config generation.

This change also removes the redundant generate_assessment_config function and the functions only it calls, because this is a completely separate flow to the thing we actually use (generate_assessment_config_for_round) in the application.

### 🚧 Testing

- Unit tests were previously completely missing for the function actually used by the application - `generate_assessment_config_for_round`
- We previously only tested the function `generate_assessment_config` and the functions it called - but this function was unused by the application, a completely separate piece of logic to `generate_assessment_config_for_round`, and we've removed it now
- Therefore this PR contains a completely rewritten test module that actually tests the thing that needs to be tested

### DIFF
--- a/app/export_config/generate_assessment_config.py
+++ b/app/export_config/generate_assessment_config.py
@@ -1,196 +1,19 @@
 import copy
 import json
-import os
-
-import click
 
 from app.all_questions.metadata_utils import form_json_to_assessment_display_types
 from app.db import db
-from app.db.models import Component, Criteria, Section, Subcriteria, Theme
-from app.db.models.application_config import READ_ONLY_COMPONENTS
-from app.db.queries.application import get_form_for_component
+from app.db.models import Form, Section
+from app.db.models.application_config import READ_ONLY_COMPONENTS, ComponentType
 from app.export_config import helpers
-from app.shared.helpers import human_to_kebab_case
+from app.shared.helpers import find_enum, human_to_kebab_case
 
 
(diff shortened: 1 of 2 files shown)


israr-ulhaq (approved):
Approving as our main goal is achieved - transitioning from ORM to JSON access for assessment config generation ✅

PR #443, opened 2025-08-19
Resolve ambiguous locators in E2E tests causing intermittent failures

wjrm500:
E2E tests were failing intermittently with Playwright strict mode violations when randomly generated section names (like "at") caused `get_by_role("heading", name=section_name)` locators to match both the intended section heading "2. at" and the main page heading "Build application" (which contains "at"). This fix updates all section-related locators to use specific CSS selectors targeting the task list structure instead of generic heading role lookups, eliminating the ambiguity and making the tests more robust.


nuwan-samarasinghe (approved):
nice change

PR #449, opened 2025-08-20
FLS-1470 - Clean up application code

wjrm500:
### 🎫 Ticket

[Clean up application code](https://mhclgdigital.atlassian.net/browse/FLS-1470)

### 🌍 Background

After completing all of the previous subtasks in this task, we’ll be in a situation where we have completely removed our functional dependency on the old code that relied on forms being stored in a relational way. Instead, FAB will be using forms stored simply as whole JSON blobs for all of its functional requirements. But we won’t have yet removed the redundant code because there is so much of it to clean up, including test code, and we didn’t want to pollute previous PRs.

Now is the time to clear out dead code. And in this ticket specifically, that does not include the “data layer” - SQLAlchemy models or real database entities via migration code. The focus here will be on all of the now-redundant stuff in e.g., `queries.py` files, and much of the stuff in both the `import_config` and `export_config` directories. We’re talking potentially thousands of lines of now-dead code, if the [spike PR](https://github.com/communitiesuk/funding-service-design-fund-application-builder/pull/429/files) is anything to go by. Basically anything that previously sat behind either `insert_form_config` (import) or `build_form_json` (export), or anything that makes reference to one of the SQLAlchemy models we’re about to dump (in [this ticket](https://mhclgdigital.atlassian.net/browse/FLS-1471)): `FormSection`, `Page`, `Lizt`, `Condition`, `PageCondition` and `Component` (from the `app.db.models.application_config` module); and `Criteria`, `Subcriteria` and `Theme` (from the `app.db.models.assessment_config` module). Then we should refactor where sensible.

### ♻️ Changes

- **Removing redundant `download_form_json` endpoint**
- **Removing `generate_form` module** - as `_find_page_by_controller` was being used in the `view_form_questions` endpoint, we just moved the logic into the endpoint itself and converted from relational access to dict access
- **Cleaning out application queries module and tests** - any reference to deprecated lower-than-form-level entities like `Page` or `Component` have been removed from the `queries.py` module. We no longer need to pass a `cascade` argument to any delete form functions as there is nothing to cascade to. These changes also required us to remove a bunch of object instantiations in the seed test data module in order to avoid foreign key constraint violations
- **Removing `test_integration` module as no longer relevant** - why is it no longer relevant? Because it was built to test that forms weren't corrupted between import and export, which made sense given the complexity of the de-construction / re-construction logic. However, given that we are now simply storing the form JSON as-is, validating this fact in painstaking detail feels like extra test code we just don't need
- **Calling `insert_new_form` directly in `create_template` endpoint** - now that we've moved all of the code around breaking down JSON forms to store as relational entities, the layers of abstraction between the template creation endpoint and the insertion of the form into the database are very much unnecessary. Currently we have `create_template` -> `json_import` -> `load_json_from_file` -> `insert_form_as_template` -> `insert_new_form`, when all we really need is `create_template` -> `insert_new_form`. This commit achieves that, but leaves the services and `load_form_json` modules untouched as they are still used by the `edit_template` endpoint
- **Fixing up `edit_template` endpoint** - now that we are moving from storing forms relationally to as whole JSON objects, we no longer need the current convoluted upload process, whereby if we get a new form JSON uploaded, we delete the existing form and import an entirely new one. Instead, we can update the form in-place using the existing `update_form` interface - we just need to pass `form_json` to it
- **Moving test-specific functions from `import_config` package into `conftest.py`, then removing now redundant `import_config` package**
- **Removing redundant `unit_test_data` module which builds complex mocks of deprecated models**
- **Removing redundant `delete_all_related_objects` calls from `delete_fund` and `delete_round` query functions**


hamzabinkhalid (approved):
Looks great to me 👍 Massive cleanup, and it’s nice to see all the legacy relational stuff finally gone. The refactors make sense and everything feels much leaner now. Appreciate the clear breakdown of changes too, makes reviewing way easier. Nice work!

PR #451, opened 2025-08-20
FLS-1470 - Remove unused ORM helper function

israr-ulhaq:
🎫 Ticket
[Clean up application code](https://mhclgdigital.atlassian.net/browse/FLS-1470)

### Change description
The get_all_pages_in_parent_form function in app/shared/helpers.py is no longer needed after the transition to using form_json. This function was querying the old Page ORM model directly, but with form_json, pages are accessed directly from the JSON structure instead.

#### Changes:

  - Remove get_all_pages_in_parent_form function
  - Remove unused from app.db.models import Page import

The function is not used anywhere in the codebase and has no associated tests, making it safe to remove as part of the  cleanup.

### DIFF
--- a/app/shared/helpers.py
+++ b/app/shared/helpers.py
@@ -2,8 +2,6 @@ from dataclasses import asdict, is_dataclass
 
 from flask import flash, render_template
 
-from app.db.models import Page
-
 
 def convert_to_dict(obj):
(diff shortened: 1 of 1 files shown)


wjrm500 (approved):
Nice catch

PR #452, opened 2025-08-20
FLS-1471 cleanup obsolete models

israr-ulhaq:
### Ticket

[Clean up data layer inc. SQLAlchemy models and database
](https://mhclgdigital.atlassian.net/browse/FLS-1471)

Completes the migration from relational form storage to JSON blob storage by removing obsolete SQLAlchemy models and database tables that are no longer needed.

### Models Removed

`FormSection, Page, Lizt, Condition, PageCondition, Component from application_config.py
`
 Entire `assessment_config.py` module (Criteria, Subcriteria, Theme)

#### Models Updated
**Form**: Removed pages and conditions relationships, `updated __repr__(`) and `as_dict()` methods
**Round**: Removed criteria relationship

Updated imports in `__init__.py` and dependent files

Preserved
ComponentType enum and READ_ONLY_COMPONENTS (required for assessment config generation)

Database Migration
Added migration to drop obsolete tables: component, page, condition, lizt, criteria, subcriteria, theme, form_section

Migration handles foreign key dependencies in correct order


app/db/migrations/versions/~2025_08_20_1631-fd9a57c67f86_remove_obsolete_relational_form_storage_.py
```diff
@@ -0,0 +1,233 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    # Drop tables in correct order to handle foreign key dependencies
+    op.drop_table("component")
+    op.drop_table("page_condition")
+    with op.batch_alter_table("page", schema=None) as batch_op:
```
wjrm500, line 25: We don't _need_ to drop the index separately if we're dropping the table - see https://www.postgresql.org/docs/current/sql-droptable.html where it says "DROP TABLE always removes any indexes, rules, triggers, and constraints that exist for the target table."

app/db/migrations/versions/~2025_08_20_1631-fd9a57c67f86_remove_obsolete_relational_form_storage_.py
```diff
@@ -0,0 +1,233 @@
...
+        batch_op.drop_index("ix_template_page_name", postgresql_where="(is_template = true)")
+    op.drop_table("page")
+    op.drop_table("condition")
+    op.drop_table("theme")
+    op.drop_table("subcriteria")
+    op.drop_table("criteria")
+    op.drop_table("lizt")
+    op.drop_table("form_section")
+    with op.batch_alter_table("round", schema=None) as batch_op:
```
wjrm500, line 34: I think this bit is for an unrelated thing, that gets auto-generated - I had this when I created the last migration, and I just manually removed it 🫨 - probably best to just remove this here as well and we can work out what it is and potentially spin up a separate PR at a later point

app/db/migrations/versions/~2025_08_20_1631-fd9a57c67f86_remove_obsolete_relational_form_storage_.py
```diff
@@ -0,0 +1,233 @@
...
+            "section_base_path",
+            existing_type=sa.INTEGER(),
+            server_default=sa.text("nextval('section_base_path_seq')"),
+            existing_nullable=True,
+        )
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
```
wjrm500, line 45: Oh my, this `downgrade` is a bit intense isn't it 😆

Going to avert my eyes...

wjrm500 (approved):
Great stuff, I've left a couple _very_ minor comments but really there isn't much to say about the code itself - a job well done! Looking forward to getting this deployed

Points of improvement? Regarding the PR description...

- It would have been good to see a bit more attention given to the formatting - we're not using code spans correctly, e.g., we have this code span `FormSection, Page, Lizt, Condition, PageCondition, Component from application_config.py` which should be `FormSection`, `Page`, `Lizt`, `Condition`, `PageCondition`, `Component` from `application_config.py`. We also have heading formatting used inconsistently e.g., "Models Removed" and "Models Updated" are different sizes, while "Preserved" and "Database Migration" feel like headings but are not formatted as such. Formatting PR descriptions is pretty unimportant in the grand scheme of things but it's nice as a reviewer to come into a PR, see that the developer has made an effort to create an attractive description, and know that you're in safe hands 🥂
- To really go above and beyond, doing some thinking about what happens next and then briefly verbalising that in the description would be amazing. For example, you could say something like "Once this PR gets merged, the migration will run automatically as part of the deployment", or mention the window between the database migration running (and deleting the tables) and the code being deployed, and consider whether there is any risk of errors occurring during that window. Or, as this is the final subtask in the Jira task, you could have a think about how we close off this task with confidence - do we need to do a final audit of the code, what would this look like, or should we actually do some manual QA in deployed environments? (_I should stress again, all of the stuff in this bullet point would very much count as going above and beyond expectations_)

wjrm500 (changes requested):
Just holding this while we investigate what happened when we deployed this branch to Dev - said "No migrations to apply", seems to suggest we are on revision `e553f67592fa` which means `0cec5f1e34b8` (make `Form.form_json` non-nullable) hasn't been applied there? See screenshot below:

Image

PR #454, opened 2025-09-03
Bugfix - remove 'name' as a required top-level field in uploaded forms

wjrm500:
Forms downloaded from Form Designer do not / do not always have this field. It doesn't look like we do anything with the value of this field, and shouldn't expect it if we don't need it.


israr-ulhaq (approved):
Fast one 👍

sc15zs (approved):
Checked out the branch and tested. 
It successfully uploads the template without the name property defined.

PR #455, opened 2025-09-12
Adding created_at and updated_at properties to Fund, Round and Form models

sc15zs:
**Ticket:**
https://mhclgdigital.atlassian.net/browse/FLS-1443

**Description:**
This PR adds `created_at` and `updated_at` properties to the following models:
- fund (displayed as “Grant” on the frontend)
- round (displayed as “Application” on the frontend)
- form (displayed as “Template” on the frontend)

The addition of `created_by` and `updated_by` properties has also been considered, but they are not included in this PR. 
For auditing purposes, adding `updated_by` would only be useful if we also tracked which property of a model was changed, along with its original and updated values. 
Since these details would not be displayed on the frontend, this PR focuses only on properties that are visible to users. 
If it’s later decided that `created_by` and `updated_by` should be stored in the database, that can be handled in a separate piece of work.

For the `created_at` property, the implementation is:
`created_at = Column(DateTime, nullable=True)`
`server_default=func.now() `was intentionally not used, because running the migration would apply the migration timestamp to all existing records, resulting in incorrect values being stored.

Instead, `created_at` is set in the application code when creating new instances of funds, rounds, and forms.


wjrm500 (approved):
Hi @sc15zs, all code makes sense and testing locally demonstrates this working correctly for all three models. Two things to mention, but neither are blockers:

- Your migration file incorrectly includes reference to `created_by` and `updated_by` in both the filename and the first line of the module-level docstring, where it should say `created_at` and `updated_at`
- Your migration file includes a reference to a column `section_base_path` in both the `upgrade` and `downgrade` functions. This thing gets added automatically when you auto-generate an Alembic migration - it must be that there is a small discrepancy between our SQLAlchemy models and the database, that we've previously ignored by manually removing the code from the migration files (maybe not the best idea, but hey-ho!). It would be good to see some information given around this in the PR description so we know why this has been added to the migration, and what the effects will be if any

Thanks

PR #457, opened 2025-09-17
Bugfix - strip leading numbers

wjrm500:
This function was faulty. If for example you had a section name 'Section1', it would iterate over all of the characters, see the '1' at the end, and then lop the first character from the start of the string (the 'S'), giving 'ection1'. Not ideal!

The consequences of this function being broken? It was only used in 'view questions' HTML pages, and from inspecting the data in production, we can see that we've fortunately not had any section names that have included digits or full stops after the initial prefix, so this will have had no effect on users.

### DIFF
--- a/app/all_questions/read_forms.py
+++ b/app/all_questions/read_forms.py
@@ -1,3 +1,6 @@
+import re
+
+
 def determine_display_value_for_condition(
     condition_value: str,
     list_name: str = None,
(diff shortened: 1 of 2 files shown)


wjrm500:
Thanks @sc15zs you are correct, in the example I gave  -"Section 1" - the old and new implementations would indeed give the same result - my mistake. The section name I tried that actually flagged the issue to me was "Section1", which the old function implementation alters to "ection1". I then wrongly extrapolated from that that "Section 1" would give "ection 1", figuring that "Section 1" would be a better example to give in the PR description.

PR #460, opened 2025-09-23
FLS-1453 - Use Pre-Award forms API

wjrm500:
### 🎫 Ticket

[FAB - Remove "Upload template" and pull forms from Pre-Award API](https://mhclgdigital.atlassian.net/browse/FLS-1453)

### 🌍 Background

We are centralising forms storage across live services.

Currently we have forms created in Form Designer and stored in the Redis cache associated with the Form Runner; these are exported from Designer and uploaded as templates in FAB; forms are cloned in FAB's internal database when they are added to an application tasklist; they are then exported from FAB as part of a bundle of configuration files; a developer will take the forms from this bundle and manually add them to the `fsd_config/form_jsons` directory of the `digital-form-builder-adapter` repo; these forms are loaded into the Redis cache when the Runner starts up.

We are re-architecting our system to support the storage of forms in a single place - a new `form_definition` table in the Pre-Award database. A new API has been created in the `funding-service-pre-award` repo to mediate data import and export to and from that table. Form Designer, Form Runner and FAB are the three separate services that will consume this API. We already have separate PRs up in the `digital-form-builder-adapter` repo that will allow Form Designer and Form Runner to use the API. This PR allows FAB to use the API.

### ♻️ Changes

This PR includes all of the changes necessary to transition FAB away from template upload and management and to Pre-Award API usage. FAB will stop storing forms as first-class entities and instead as links between application tasklists and forms stored in Pre-Award.

FAB will use the Pre-Award API for a number of different use cases:

- To list forms that can be added to an application tasklist in a drop-down selector on the "Edit section" page
- To get form JSON configurations for generating HTML representations of both individual forms and entire application tasklists
- To get form JSON configurations for generating assessment config
- To get form JSON configurations for preview\*
- To get form JSON configurations for export\*\*
- To get form display names for rendering a reasonable facsimile of the application tasklist
- To get form display names for generating application config
- To get form display names needed for generating assessment config 

\* This is deprecated, but we are keeping it for now so that we can deploy this PR in isolation. Basically, with Form Runner in its current form, we need to publish a form JSON configuration to it in order to have that form rendered. In future, Form Runner will pull from the same data source as Designer and FAB, making publishing redundant.

\*\* This is also deprecated, again we're keeping it to allow us to deploy this PR in isolation and not break any workflows. In future, we'll not need to export forms from FAB at all - they'll all be directly available in the Form Runner via the API.

Here is a list of the commits in this branch:

- Remove redundant unit test setup methods `load_form_jsons` and `db_with_templates`
- Remove entire templates section from FAB
- Add Pre-Award API service layer
- Add `url_path` column to Form model
- Switch form selection during build application to use Pre-Award API
- Insert new forms during round cloning instead of cloning forms
- Replace `insert_new_form` with `insert_new_section_form` in tests and remove `insert_new_form`
- Rename `insert_new_section_form` to simpler `insert_form`
- Remove extraneous fields from Form model
- Remove `name_in_apply_json` from Form model
- Add universal form store API service method for getting display name from URL path to reduce duplicate code
- Remove `form_json` from Form model
- Remove `runner_publish_name` from form table
- Allow PR-scoped E2E tests to pass with a stub server

Please read individual commit descriptions for more detail.

### 📸 Screenshots

There isn't too much to show here, other than the home page which no longer includes template-related content or the "Templates" link in the navigation bar:

image


hamzabinkhalid (approved):
Went through all commits and changes with @wjrm500 , everything is quite clear and works as expected, very good work :-)

PR #480, opened 2026-08-03
BAU: Fix uv caching cleanup bug in workflows

MarcUsher:
There seems to be a known post-job cache-pruning bug with uv setup, specifically when running inside a container like the microsoft one we have it in here.

Common fix seems to be disabling the cache clean-up, which for a legacy repo seems like an OK low-effort fix in order to unblock the pipeline.


samuelhwilliams:
Example failure: https://github.com/communitiesuk/funding-service-design-fund-application-builder/actions/runs/30813269464/job/91691974610

samuelhwilliams (approved):

